PR #32, opened 2014-10-11
Inclusion of Hclust into Clustering.jl

davidavdav:
Hi, 

Following the suggestions from the METATDATA maintainers, I made a patch for Clustering to absorb the package Hclust (Hierarchical clustering along the lines of R's hclust() and cutree()), sice this is also does a form of clustering.  This keeps the number of packages down.  

I did not further integrate the code, it seems to be reasonably independent of the existing code in Clustering. 

---david

### DIFF
--- a/src/hclust.jl
+++ b/src/hclust.jl
@@ -1,0 +1,388 @@
+## hclust.jl (c) 2014 David A. van Leeuwen 
+## Hierarchical clustering, similar to R's hclust()
+
+## Algorithms are based upon C. F. Olson, Parallel Computing 21 (1995) 1313--1325. 
+
+## This is also in types.jl, but that is not read...
+## Mostly following R's hclust class
+type Hclust{T<:Real}
+    merge::Matrix{Int}
+    height::Vector{T}
+    order::Vector{Int}
+    labels::Vector
+    method::Symbol
+end
+
+function assertdistancematrix(d::AbstractMatrix)
+    nr, nc = size(d)
+    nr == nc || throw(DimensionMismatch("Distance matrix should be square."))
+    issym(d) || error("Distance matrix should be symmetric.")
+end
+
+## This seems to work like R's implementation, but it is extremely inefficient
+## This probably scales O(n^3) or worse. We can use it to check correctness
+function hclust_n3{T<:Real}(d::AbstractMatrix{T}, method::Function)
+    assertdistancematrix(d)
+    mr = Int[]                  # min row
+    mc = Int[]                  # min col
+    h = T[]                     # height
+    nc = size(d,1)              # number of clusters
+    cl = -[1:nc]                # segment to cluster attribution, initially negative
+    next = 1                    # next cluster label
+    while next < nc 
+        mindist = Inf
+        mi = mj = 0
+        cli = unique(cl)
+        mask = BitVector(nc)
+        for j in 1:length(cli)           # loop over for lower triangular indices, i>j
+            cols = cl .== cli[j]
+            for i in (j+1):length(cli)
+                rows = cl.==cli[i]
+                distance = method(d[rows,cols]) # very expensive
+                if distance < mindist
+                    mindist = distance
+                    mi = cli[i]
+                    mj = cli[j]
+                    mask = cols | rows
+                end
+            end
+        end 
+        ## simulate R's order
+        if mi < 0 && mj < 0 && mi > mj ||
+            mi > 0 && mj > 0 && mi < mj ||
+            mi < 0 && mj > 0
+            push!(mr, mi)
+            push!(mc, mj)
+        else
+            push!(mr, mj)
+            push!(mc, mi)
+        end
+        push!(h, mindist)
+        cl[mask] = next
+        next += 1
+    end
+    hcat(mr, mc), h
+end
+
+## Efficient single link algorithm, according to Olson, O(n^2), fig 2. 
+## Verified against R's implementation, correct, and about 2.5 x faster
+## For each i < j compute D(i,j) (this is already given)
+## For each 0 < i <= n compute Nearest Neighbor N(i)
+## Repeat n-1 times
+##   find i,j that minimize D(i,j)
+##   merge clusters i and j
+##   update D(i,j) and N(i) accordingly
+function hclust_minimum{T<:Real}(ds::Symmetric{T})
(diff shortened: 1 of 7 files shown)


davidavdav:
Anything else needed for considering this PR?

src/hclust.jl
```diff
@@ -0,0 +1,452 @@
...
+## Efficient single link algorithm, according to Olson, O(n^2), fig 2. 
+## Verified against R's implementation, correct, and about 2.5 x faster
+## For each i < j compute D(i,j) (this is already given)
+## For each 0 < i <= n compute Nearest Neighbor N(i)
+## Repeat n-1 times
+##   find i,j that minimize D(i,j)
+##   merge clusters i and j
+##   update D(i,j) and N(i) accordingly
+function hclust_minimum{T<:Real}(d::Matrix{T})
+    @assert size(d,1) == size(d,2)
```
jpfairbanks: Should this be the more idiomatic:
`size(d,1) == size(d,2) || error("Distance matrix d is not square.")`?
That error might be more descriptive than the output of the failed assert.

src/hclust.jl
```diff
@@ -0,0 +1,452 @@
...
+    ## For each i < j compute d[i,j] (this is already given) --- we need a local copy
+    if !issym(d)
```
jpfairbanks: Maybe auto-symmetrizing the matrix is suboptimal. I might want to use `d = triu(d) + triu(d)'` or `d = tril(d)+tril(d)'` and if I forget to do that before passing in `d`, then it might be harder to track down that bug if we auto-symmetrize with the `d+=d'`.

we could use the line `issym(d) || error("Distance matrix d is not symmetric` to test and error.

src/hclust.jl
```diff
@@ -0,0 +1,452 @@
...
+type Hclust{T<:Real}
+    merge::Matrix{Int}
+    height::Vector{T}
+    order::Vector{Int}
+    labels::Vector
+    method::Symbol
+end
+
+function hclust_n3{T<:Real}(d::Matrix{T}, method::Function)
+    @assert size(d,1) == size(d,2)
```
jpfairbanks: Same thing as above about the error message instead of assert. and auto symmetrization.

jpfairbanks:
I would like to see this merged, one thing that I can see is that you use built in function names as variable names. specifically you assign to `min` and `new`. It leads to a possible confusion. 

The functions are also quite large. I think there are some pieces of functionality that would be better made into functions that modify their arguments that get called more than once. Refactoring it so that these blocks become reusable pieces of code will help review the PR.

For instance:
````julia
## For each 0 < i <= n compute Nearest Neighbor N[i]
    N = zeros(Int, nc)
    for k = 1:nc
        min = Inf
        mk = 0
        for i = 1:(k-1)
            if d[i,k] < min
                min = d[i,k]
                mk = i
            end
        end
        for j = (k+1):nc
            if d[k,j] < min
                min = d[k,j]
                mk = j
            end
        end
        N[k] = mk
    end
````
could be extracted into a function such as `nearestneighbors!(N, d, nc)`.

````julia
## For each 0 < i <= n compute Nearest Neighbor N[i]
function nearestneighbors!(N,d,nc)
    for k = 1:nc
        min = Inf
        mk = 0
        for i = 1:(k-1)
            if d[i,k] < min
                min = d[i,k]
                mk = i
            end
        end
        for j = (k+1):nc
            if d[k,j] < min
                min = d[k,j]
                mk = j
            end
        end
        N[k] = mk
    end
end
#usage
N = zeros(Int, nc)
nearestneighbors!(N,d,nc)
````

File: src/hclust.jl (hunk 2 above)
Comment (davidavdav) [inline src/hclust.jl]: I am not too happy with issym() myself as it takes some time.

Comment (jpfairbanks) [inline src/hclust.jl]: We could just take a lower trianglular matrix. It would be the users
responsibility to ensure that the distance function they computed was
symmetric. Taking the triangular input also saves some time for the user.

Comment (davidavdav) [inline src/hclust.jl]: I'd be happiest with a matrix type Symmetric, but we can't really work with that because setindex!() is not defined for Symmetric.

**Comment by davidavdav** [comment]
I renamed min and max, and got some common code out of the functions.  I agree that the functions are quite long, especially hclust_minimum() (single linkage clustering).  This is a bit frustrating, as the pseudocode is only 6 lines...  I don't see a way to make re-usable code out of it.

**Comment by davidavdav** [comment]
The interface now is 
```julia
hclust(d::Symmetric, method::Symbol) ## method :single, :complete, or :average
```
with convenience functions
```julia
hclust(d::AbstractMatrix, method::Symbol) ## checks symmetry
hclust(d::AbstractMatrix, method::Symbol, uplo) ## use only upper/lower part of d
```

In the last function, `uplo` is a Char (for Julia v0.3) or a Symbol (for Julia v0.4). 

I checked the speed of access to elements in a `Symmetric`, it appears to be as fast as a normal matrix, perhaps even faster because more elements fit in the cache.

jpfairbanks:
@davidavdav Why do you want to include hclust1 if it doesn't work? Is the intended algorithm supposed to be faster or have better statistical properties when coded correctly?

davidavdav:
Maybe someone else can pick it up.  It may be a more efficient algorithm than hclust2()---that still is 4.5x as slow as R's implentation (where hclust_minimum() is faster than R). 

But there is no ham in removing it.

davidavdav:
Does anything still needs to be done for this PR?

slundberg:
Just checking on the status on this pull request. I want to release code that depends on HClust so this would effect how I incorporate the dependency.

johnmyleswhite:
Just to confirm: you've never read the code for R's implementation, right? Assuming that's true, I'd be ready to squash this and merge it.

davidavdav:
I did not look in R's code for hclust---I wouldn't even know where to find it.  I used the paper quoted in the source (Olson).  I couldn't get the Wikipedia algorithm description working.

johnmyleswhite:
Works for me. Thanks for writing this!

johnmyleswhite:
Can you do a quick squash? Then we'll merge this.

davidavdav:
I don't think the travis 0.4 failure is related to this squashed commit.

johnmyleswhite:
Sounds like the code is broken on 0.4. We can fix that.

tlnagy:
Any progress on this?

PR #34, opened 2014-12-31
WIP: PCCA+

axsk:
Hey,
this is an implementation of Robust Perron Cluster Cluster Analysis (PCCA+),
an algorithm clustering metastable conformations in Markov Chains,
and before writing further tests and docs I wanted to ask how high my chances are getting this added to Clustering.jl.

I am also working on an module for spatial timeseries clustering (using PCCA+), but I guess this is too specific to be added here?

Best regards
axsk

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,3 +2,4 @@ julia 0.3-
 ArrayViews 0.4.8-
 Distances 0.2-
 StatsBase 0.6.9-
+Optim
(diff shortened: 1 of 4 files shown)


axsk:
I could do that. Actually the algorithm computes a (in some cases sufficient) start-clustering and then optimizes from there on, so the I could make the whole Optim part optional.
But I couldn't find out yet how to handle optional dependencies in Julia.

axsk:
Closed in favor of the [PCCAPlus.jl](https://github.com/axsk/PCCAPlus.jl) package

PR #39, opened 2015-02-24
Extend varinfo to include other normalized variants.

jpfairbanks:
Variation of information can be normalized in different ways based on the properties of the application.
I added a symbol argument to specify this normalization to the `varinfo` function and set the default to the previous behavior. I also put the computation of `I, H1, H2` into a function `information` so that it could be reused in the future. This is a nonbreaking change.

If there is a better way to do it, let me know and I can change the pull request. I think this is the least disruptive change possible. I added tests and documentation to match what was there.

The paper cited is here
http://jmlr.csail.mit.edu/papers/volume11/vinh10a/vinh10a.pdf

### DIFF
--- a/doc/source/varinfo.rst
+++ b/doc/source/varinfo.rst
@@ -1,7 +1,7 @@
 Variation of Information
 ==========================
 
-`Variation of information <http://en.wikipedia.org/wiki/Variation_of_information>`_ (also known as *shared information distance*) is a measure of the distance between two clusterings. It is devised based on mutual information, but it is a true metric, *i.e.* it satisfies symmetry and triangle inequality. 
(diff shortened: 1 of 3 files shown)


pcjentsch:
This PR would be useful to me, what prevented it from being merged?

The same authors that @jpfairbanks cites have since published a [follow-up paper](https://jmlr.csail.mit.edu/papers/volume11/vinh10a/vinh10a.pdf), which brought me to this PR.

alyst:
@pcjentsch I guess it was a lack of maintainers resources :( Unfortunately, the PR was submitted quite long time ag, and the codebase has changed since. To the level that there are merge conflicts, and it cannot be merged automatically. If @jpfairbanks would be so kind to review and rebase his work, I would be happy to review it. Alternatively ‐ if he would be ok with that - you can fork this PR and submit an updated version.

PR #44, opened 2015-06-17
fixed bug in dbscan _dbs_expand_cluster! which would create repeated …

wolffmb:
…elements in the neighbor array

qnbs in db_expand_cluster compiles a list of neighbors for q. this list is then merged with the nbs array to track all neighbors. However, arrays allow duplicate elements, so if an elements was in qnbs and in nbs, the element would still be pushed onto nbs, creating duplicate elements in nbs.


alyst:
closing as outdated

PR #46, opened 2015-07-01
change julia require from 0.3- to 0.3

tkelman:
assuming no one is continuing to test with prerelease versions of 0.3


johnmyleswhite:
Thanks!

PR #45, opened 2015-06-30
Added unit tests and an assert to cover a bounds error on small examples

slundberg:
This adds Travis unit tests for the hclust implementation, and also an assert that more than 3 variables are being clustered since the current algorithm assumes that.

I also updated a tuple type annotation in kmedoids.jl to address a Travis error in v0.4

PR #47, opened 2015-07-02
Fix bug found by @slundberg in #45 for small distance matrices

davidavdav:
Scott Lundberg found a bug in `hclust2()` apparent with low probability for smaller input matrices.  This PR fixes the bug which is the result of a too small array used for a stack.  The test script now extends to matrices as small as 2x2.


slundberg:
Anything else needed to get this merged?

Issue #48, opened 2015-07-08
New validation measures


jumutc:
Hello folks,

What about new validation measures for clustering like: Rand index, Adjusted Rand index etc.?
Please take a look at this [Wikipage](https://en.wikipedia.org/wiki/Rand_index) and corresponding publication in [Journal of Classification](http://link.springer.com/article/10.1007/BF01908075).

johnmyleswhite:
Pretty busy these days, but I'll try to review your PR in a few weeks.

PR #49, opened 2015-07-08
Implementation for issue #48

jumutc:
I have taken and refactored a bit implementation for issue https://github.com/JuliaStats/Clustering.jl/issues/48 from the [MATLAB](http://www.mathworks.com/matlabcentral/fileexchange/13916--simple--tool-for-estimating-the-number-of-clusters/content/valid_RandIndex.m) central

### DIFF
--- a/doc/source/randindex.rst
+++ b/doc/source/randindex.rst
@@ -1,0 +1,33 @@
+Rand indices
+==========================
+
+`Rand index <http://en.wikipedia.org/wiki/Rand_index>`_ is a measure of the similarity between two data clusterings. From a mathematical standpoint, Rand index is related to the accuracy, but is applicable even when class labels are not used.
+
+**References:**
+
+    Lawrence Hubert and Phipps Arabie (1985). 
+    *Comparing partitions.* 
+    Journal of Classification 2 (1): 193–218 
+
+    Meila, Marina (2003). 
+    *Comparing Clusterings by the Variation of Information.* 
+    Learning Theory and Kernel Machines: 173–187.
+
+This package provides the ``randindex`` function that implements several metrics:
+
+.. function:: randindex(c1, c2)
+
+    Compute the tuple of indices (Adjusted Rand index, Rand index, Mirkin's index, Hubert's index) between two assignments. 
+
+    :param c1: The assignment vector for the first clustering.
+    :param c2: The assignment vector for the second clustering.
+
+    :return: tuple of indices.
+
+.. function:: varinfo(R, k0, a0)
+
+    This method takes ``R``, an instance of ``ClusteringResult``, as input, and computes the tuple of indices (see above) where ``a0`` is the corresponding assignment vector. 
+
+.. function:: varinfo(R1, R2)
+
+    This method takes ``R1`` and ``R2`` (both are instances of ``ClusteringResult``) and computes  the tuple of indices (see above) between them.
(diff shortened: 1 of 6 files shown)


doc/source/randindex.rst
```diff
@@ -0,0 +1,33 @@
+Rand indices
+==========================
+
+`Rand index <http://en.wikipedia.org/wiki/Rand_index>`_ is a measure of the similarity between two data clusterings. From a mathematical standpoint, Rand index is related to the accuracy, but is applicable even when class labels are not used.
+
```
jpfairbanks: What does "applicable even when class labels are not used" mean? Does it mean even when the clustering labels do not partition the data, or when there are more than 2 classes? In our case we have "class labels" its the identifier of the cluster to which the data belongs. I looked at the wiki article and was confused by it there too.

src/randindex.jl
```diff
@@ -0,0 +1,59 @@
...
+	#
+
+	contigency = zeros(maximum(mem1),maximum(mem2))
+
+	for i = 1:length(mem1)
+	   contigency[mem1[i],mem2[i]] += 1
+	end
+
+	return contigency
+end
```
jpfairbanks: Should this be replaced by a call to StatsBase.counts? It is already imported above.

jpfairbanks:
going though the math I was a a little confused comparing to the formula on the wikipedia page linked to in the source.

It appears simpler to write something like:
```julia
A = 0
for i in 1:size(c,1)
    for j in 1:size(c,2)
        A += binom(c[i,j], 2)
    end
end

a = sum(c,1)
b = sum(c,2)
nc = 0
abinoms = [binom(a[i],2) for i in 1:size(c,1)]
bbinoms = [binom(b[j],2) for j in 1:size(c,2)]

for i in size(c,1)
    for j in size(c,2)
        nc += abinoms[i] * bbinoms[j]
    end
end
nc = nc / binom(n,2)
```

Am I missing something important?

doc/source/randindex.rst
jumutc: Rand index applies to the case when the distribution of classes/memberships per assignment is different. So index relies on the assignment as in the case of ```varinfo```. Wiki article is a little bit confusing but the definition of Rand index is correct from my point of view.

jumutc:
I think this implementation of Rand-related indices is more or less optimal. Take into account multiple measures we have to address (not only Rand index itself). This implementation is tested quite well by other researchers as well so I tend to accept it and rely on it as it is.

johnmyleswhite:
Just remembered that the use of code from Matlab Central is a problem for us. As Tony Kelman recently noted in another context:

> Note that there's the awful "Mathworks products only" clause in the File Exchange terms of use, see 2.a.iii. at http://www.mathworks.com/matlabcentral/termsofuse.html#content. So despite the BSD license on the code, unless it's available through a different medium or directly from the author, having looked at any code hosted on File Exchange means we can't distribute a translation of it in Julia.

jpfairbanks:
So to escape this clause, can you find code on Matlab Central and then ask
the original author for it? What if the code is also posted on a different
website by the original author?

johnmyleswhite:
Judging from the conversation on Base about the log of a matrix, I gather that people feel that the author's code (if posted elsewhere) is usable. But you may also need permission from the author.

jumutc:
I have asked for permission. Waiting for response...

dcorney:
Hi - I'm the original author of the Matlab code used. I'm very happy (and flattered!) for you to use this as you see fit.
Best wishes - David

PS And thanks for asking!

johnmyleswhite:
Thanks all for doing this!

PR #59, opened 2016-02-10
Implemented order attribute for hclust.

slundberg:
Before the order attribute was just a placeholder. This implements it and adds it to the testing file to ensure it produces the same results as the R code. (I didn't look at R when writing the few lines to implement this)


ahwillia:
+1, I'd like to use this.

slundberg:
Any status on this? Thanks!

diegozea:
@kmsquire  this is fine :)
The change it's simple and gives the same order than R's `hclust`
i.e.:
```julia
julia> for i in 1:10
           ie = Symmetric(rand(100,100))
           bool = hclust(ie, :single).order .== rcopy(R"""hclust(as.dist($ie), method="single")""")[:order]
           println(mean(bool))
       end
1.0
1.0
1.0
1.0
1.0
1.0
1.0
1.0
1.0
1.0

julia> for i in 1:10
           ie = Symmetric(rand(100,100))
           bool = hclust(ie, :complete).order .== rcopy(R"""hclust(as.dist($ie), method="complete")""")[:order]
           println(mean(bool))
       end
1.0
1.0
1.0
1.0
1.0
1.0
1.0
1.0
1.0
1.0

julia> for i in 1:10
           ie = Symmetric(rand(100,100))
           bool = hclust(ie, :average).order .== rcopy(R"""hclust(as.dist($ie), method="average")""")[:order]
           println(mean(bool))
       end
1.0
1.0
1.0
1.0
1.0
1.0
1.0
1.0
1.0
1.0

```

kmsquire:
Thanks @diegosea!  

I noticed that this function dispatches on a keyword argument. It might be more Julian if it used multiple dispatch (although sometimes the current method is better).

PR #61, opened 2016-02-29
PR: memory efficient DBSCAN clustering

OmriRoames:
The existing DBSCAN uses a full adjacency matrix which blows the memory when a large number of points is used. Here is an implementation that build an adjacency list on the fly. Also it reports the indices of both the core points of each cluster and the boundary points which can belong to multiple clusters.

### DIFF
--- a/src/dbscan.jl
+++ b/src/dbscan.jl
@@ -102,4 +102,86 @@ function _dbs_expand_cluster!{T<:Real}(D::DenseMatrix{T},           # distance m
     return cnt
 end
 
+using NearestNeighbors
 
+
+immutable Cluster
+  size::Int64
+  core_indices::Array{Int64, 1}
+  boundary_indices::Array{Int64, 1}
+end
+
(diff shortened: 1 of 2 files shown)


src/dbscan.jl
```diff
@@ -15,6 +17,13 @@ type DbscanResult <: ClusteringResult
 end
 
 
+immutable Cluster
+  size::Int64
```
ararslan: Please use the standard 4-space indents for consistency with the rest of the code

ararslan:
Can you rebase from master and resolve the branch conflicts?

OmriRoames:
Sure. As I mentioned, this code is not up to date. I'll update it and rebase it.

ararslan:
Definitely. This package current requires 0.4, we just don't have a tagged version with that requirement.

ararslan:
Closing in favor of #69 I suppose

PR #65, opened 2016-06-13
Add MCL (Markov Clustering) + tests

alyst:
Adds MCL (Markov Cluster Algorithm) for weighted graphs (http://micans.org/mcl/).

### DIFF
--- a/src/mcl.jl
+++ b/src/mcl.jl
@@ -121,11 +121,12 @@ function mcl(adj::Matrix{Float64};
     rel_delta = NaN
     while !converged && niter < max_iter
         # expand (raise to the matrix power)
-        next_mcl_adj = mcl_adj ^ expansion # FIXME don't reallocate
+        expanded = mcl_adj ^ expansion # FIXME don't reallocate
         # inflate (apply power element-wise)
-        @inbounds for i in eachindex(next_mcl_adj)
-            next_mcl_adj[i] ^= inflation
+        @inbounds for (i, el) in enumerate(expanded)
+            next_mcl_adj[i] = abs(el)^inflation
         end
+
         # normalize in columns
         scale!(next_mcl_adj, 1.0./squeeze(sum(next_mcl_adj, 1), 1))
 
--- a/src/mcl.jl
+++ b/src/mcl.jl
@@ -79,8 +79,21 @@ end
 
 # adjacency matrix expansion (matrix-wise raising to a given power) kernel
 # FIXME `_mcl_expand!()` that does not allocate new expanded matrix
-function _mcl_expand(src::Matrix, expansion::Number)
+function _mcl_expand(src::Matrix, expansion::Integer)
     src ^ expansion
+end
+
+# adjacency matrix expansion (matrix-wise raising to a given power) kernel
+# FIXME `_mcl_expand!()` that does not allocate new expanded matrix
+function _mcl_expand(src::Matrix, expansion::Number)
+    # we have to implement the workarond for matrix-power because of julia bug #16930
+    if isinteger(expansion)
+        return _mcl_expand(src, Integer(real(expansion)))
+    end
+    v, X = eig(src)
+    (eltype(v) <: Complex) || (any(v.<0) && (v = complex(v)))
+    Xinv = ishermitian(src) ? X' : inv(X)
+    scale(X, v.^expansion)*Xinv
 end
 
 # adjacency matrix inflation (element-wise raising to a given power) kernel
--- a/src/mcl.jl
+++ b/src/mcl.jl
@@ -1,0 +1,164 @@
+# MCL (Markov CLustering algorithm)
+
+#### Interface
+
+"""
+    `mcl()` result.
+"""
+immutable MCLResult <: ClusteringResult
+    mcl_adj::Matrix{Float64}    # final MCL adjacency matrix (equailibrium state matrix if converged)
+    assignments::Vector{Int}    # element-to-cluster assignments (n)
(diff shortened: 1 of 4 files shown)


kmsquire:
Failing on v0.3.

The easiest way forward is to drop support for v0.3.  It's a pretty simple thing to do, but I don't have time right now.  Assuming there are no disagreements, a separate PR for that would be welcome.  (Cc: @johnmyleswhite)

alyst:
0.3 support was dropped, the light is green now (also I've fixed some bigs).
It would be great if somebody could review the PR, esp. for missing docs/tests.

src/mcl.jl
```diff
@@ -0,0 +1,214 @@
+# MCL (Markov CLustering algorithm)
+
+#### Interface
+
+"""
+    `mcl()` result.
```
ararslan: Should be more descriptive and structured differently, something like
```julia
"""
    MCLResult

The type of the value returned by `mcl`.

# Fields

mcl_adj      :: Matrix{Float64}  - Final MCL adjacency matrix (equailibrium state matrix if converged)
assignments  :: Vector{Int}      - Element-to-cluster assignments (n)
counts       :: Vector{Int}      - Number of samples assigned to each cluster (k)
nunassigned  :: Int              - Number of single elements not assigned to any cluster
iterations   :: Int              - Number of elapsed iterations
rel_Δ        :: Float64          - Final relative Δ
converged    :: Bool             - Whether the procedure converged
"""
```

src/mcl.jl
```diff
@@ -0,0 +1,214 @@
...
+
+""" _mcl_clusters(mcl_adj, allow_singles, zero_tol = 1E-20)
+
+    Extract clusters from the final MCL matrix `mcl_adj`.
+    Returns the vector of cluster indices for each element, the sizes of the
+    clusters and the number of the unassigned elements (`allow_singles` controls
+    whether 1-element clusters would be retained or collapsed into the set of
+    unassigned elements with the cluster index 0).
+
+    See also `mcl()`.
```
ararslan: Unindent all lines except for the function signature. But since this function isn't exported and isn't meant to be called directly, IMO it probably shouldn't have a docstring. You could convert the docstring to comments since the description is still useful to have.

src/mcl.jl
```diff
@@ -0,0 +1,214 @@
...
+end
+
+""" mcl(adj::Matrix)
+
+    Identifies clusters in the weighted graph specified by its adjacency
+    matrix using Markov Clustering Algorithm (MCL).
+
+    Returns `MCLResult`.
+
+    See http://micans.org/mcl/
```
ararslan: Should unindent all lines except for the function signature. Also, the description of what it does should be in the imperative, e.g. "_Identify_ clusters in the weight graph" (note "identify" versus "identifies").

Rather than the line `Returns MCLResult`, you could make the function signature
```julia
    mcl(adj::Matrix{Float64}) -> MCLResult
```
Having the eltype in there as above would be helpful.

src/mcl.jl
```diff
@@ -0,0 +1,214 @@
...
+function mcl(adj::Matrix{Float64};
+             add_loops::Bool = true,
+             expansion::Number = 2, inflation::Number = 2.0,
+             save_final_matrix::Bool = false,
+             allow_singles::Bool = true,
+             max_iter = 100, tol::Number=1E-5,
+             display::Symbol=:none)
+
+    m, n = size(adj)
+    m == n || throw(ArgumentError("Square adjacency matrix expected"))
```
ararslan: Should use `DimensionMismatch` rather than `ArgumentError`

src/mcl.jl
```diff
@@ -0,0 +1,214 @@
...
+        if !isfinite(next_mcl_norm)
+            warn("MCL adjacency matrix norm is not finite")
+            break
+        end
+        rel_delta = euclidean(next_mcl_adj, mcl_adj)/mcl_norm
+        if display == :verbose
+            info("MCL iter. #$niter: rel.Δ=", rel_delta)
+        end
+        converged = rel_delta <= tol
+        if converged break end
```
ararslan: `converged && break` rather than single line `if end`

src/mcl.jl
```diff
@@ -0,0 +1,214 @@
...
+        # update (swap) MCL adjacency
+        niter += 1
+        mcl_adj, next_mcl_adj = next_mcl_adj, mcl_adj
+        mcl_norm = next_mcl_norm
+        if mcl_norm < tol break end # matrix is zero
```
ararslan: here too

src/mcl.jl
alyst: I would rather indicate that it's immutable on the first line, than write it's a type (the latter is a little bit like explicitly writing "function that <...>" when documenting a function).

I think documenting the fields out-of-line and keeping both places in sync is too much of a burden, I would rather use some docstring extension that automatically syncs inline comments. Anyway, the fields should not be accessed directly in most of the cases, the user should use `ClusteringResult` API.

alyst:
@ararslan Thanks for the review. I tried to address all the points.

alyst:
Would it be fine to merge it within the next few days?

ararslan:
Okay by me.

ararslan:
One of the MCL tests is consistently failing on 0.6 macOS, but _only_ on 0.6 macOS. The check that's being hit is [this](https://github.com/JuliaStats/Clustering.jl/blob/master/src/mcl.jl#L108). I haven't been able to figure out what it is that's different with macOS that could be causing the failure there and only there.

Issue #58, opened 2016-02-06
DBSCAN should not require a distance matrix


kno10:
Distance matrixes are too expensive.
They need O(n^2) memory, and thus O(n^2) time to create.

Better DBSCAN implementations use indexes for acceleration.

johnmyleswhite:
Thanks for the tip. Closing as a non-priority.

kmsquire:
@kno10, a pull request which implemented the indexed based method would be welcome!

kno10:
Even without an index, it would allow running the algorithm on much larger data.
By precomputing a distance matrix, you run of out memory at 50k instances, which is not too big yet.
It takes about 90 seconds on a single core in C++ to run DBSCAN on 110k instances, but Julia is out of memory at half of that data size (also affects e.g. hclust).

Don't expect a pull request from me: I've written less than 10 lines of Julia, and it already annoyed me that it uses column-major when every language I use else uses row-major order ...

simonster:
I don't really know the algorithm here and I'm not making a commitment to implement anything, but perhaps this as simple as using `inball` from NearestNeighbors.jl in place of [`_dbs_region_query`](https://github.com/JuliaStats/Clustering.jl/blob/3f1fd76d943cc8a4ab600f692cc6622d1b8fba76/src/dbscan.jl#L62-L72)?

kno10:
@simonster yes, that function is meant to retrieve all objects within a radius epsilon (including the query point)

KristofferC:
You could also look at (the package which happens to have the same name...) https://github.com/KristofferC/NearestNeighbors.jl which is quite fast.

```
KDTree(rand(3,10^6))
  1.601745 seconds (100.02 k allocations: 65.614 MB, 2.79% gc time)

@time nbs = inrange(tree, rand(3,10^6), 0.01)
  4.153109 seconds (2.41 M allocations: 170.713 MB, 14.72% gc time)
```

KristofferC:
Reopening this since it seems like it is more likely to get work done on if it is seen by more people. Doesn't mean anyone is actively working on it.

OmriRoames:
Is anyone still interested in it? I have an implementation that uses an adjacency list so it's memory efficient. It also keeps track of the core and boundary.

ararslan:
@OmriRoames I don't see why not. Pull requests are always welcome. 😄

OmriRoames:
[...]

PR #69, opened 2016-07-20
Memory efficient DBSCAN

OmriRoames:
* Uses adjacency list rather than adjacency matrix (see #58 )
* Suports PointCloud container
* keeps track of core and boundary points

Updated pull request instead of #61


src/dbscan.jl
```diff
@@ -102,4 +109,113 @@ function _dbs_expand_cluster!{T<:Real}(D::DenseMatrix{T},           # distance m
     return cnt
 end
 
+"""
+    dbscan(points, radius [; leafsize = 20, min_neighbors = 1, min_cluster_size = 1]) -> clusters
```
ararslan: No need for brackets around keyword arguments

src/dbscan.jl
```diff
@@ -15,6 +15,13 @@ type DbscanResult <: ClusteringResult
 end
 
 
+immutable DbscanCluster <: ClusteringResult
```
ararslan: IMO since DBSCAN is an acronym, it should be in all caps, i.e. `DBSCANCluster`

src/dbscan.jl
```diff
@@ -102,4 +109,113 @@ function _dbs_expand_cluster!{T<:Real}(D::DenseMatrix{T},           # distance m
...
+            _update_exploration_list!(adj_list, to_explore, visited)
+        end
+        cluster_size = sum(cluster_selection)
+        min_cluster_size <= cluster_size && _accept_cluster!(clusters, core_selection, cluster_selection, cluster_size)
+    end
+    return clusters
+end
+
+"""
+`_update_exploration_list!(adj_list, exploration_list, visited)`
```
ararslan: Should indent 4 spaces rather than use backticks in docstrings

src/dbscan.jl
```diff
@@ -102,4 +109,113 @@ function _dbs_expand_cluster!{T<:Real}(D::DenseMatrix{T},           # distance m
     return cnt
 end
 
+"""
+    dbscan(points, radius [; leafsize = 20, min_neighbors = 1, min_cluster_size = 1]) -> clusters
+
+Run DBSCAN spatial clustering for the given `points`.
```
ararslan: > Cluster `points` using the DBSCAN (density-based spatial clustering of applications with noise) algorithm.

?

REQUIRE
```diff
@@ -1,4 +1,6 @@
 julia 0.4
+Compat 0.8.5
 Distances 0.3.1
+FixedSizeArrays
 StatsBase 0.9.0
-Compat 0.8.5
+PointClouds
```
ararslan: So the dependency on PointClouds is new? I'm not familiar with it. Is it actively maintained? If it's only being used for one function (which also has another method) then it's probably not worth having another dependency.

Also should specify minimum versions for all package requirements.

ararslan: PointClouds isn't registered in METADATA, which is causing the failures.

src/dbscan.jl
```diff
@@ -102,4 +109,113 @@ function _dbs_expand_cluster!{T<:Real}(D::DenseMatrix{T},           # distance m
...
+### Output
+* `Vector{DbscanCluster}`: an array of clusters with the id, size core indices and boundary indices
+
+### Example:
+points = randn(3, 10000)
+clusters = dbscan(points, 0.05, min_neighbors = 3, min_cluster_size = 20) # clusters with less than 20 points will be discarded
+"""
+function dbscan{T <: AbstractFloat, N}(points::Array{T, N}, radius::Real; leafsize::Int = 20, kwargs ...)
+    kdtree = KDTree(points; leafsize=leafsize)
+    kdtree::KDTree{Float64, Distances.Euclidean} # TODO: julia-0.4 fails type inference here
```
KristofferC: Hmm I should fix this in NearestNeighbors...

REQUIRE
KristofferC: Need to add NearestNeighbors here as well.

src/dbscan.jl
```diff
@@ -102,4 +109,113 @@ function _dbs_expand_cluster!{T<:Real}(D::DenseMatrix{T},           # distance m
...
+    dbscan(points, radius [; leafsize = 20, min_neighbors = 1, min_cluster_size = 1]) -> clusters
+
+Run DBSCAN spatial clustering for the given `points`.
+
+### Arguments
+* `points`: matrix of points or a `PointCloud`
+* `radius::Real`: query radius
+
+### Keyword Arguments
+* `leafsize::Int`: number of leafs in the `KDTree`
```
KristofferC: It is more like the number if points binned in each leaf node.

src/dbscan.jl
```diff
@@ -102,4 +109,113 @@ function _dbs_expand_cluster!{T<:Real}(D::DenseMatrix{T},           # distance m
...
+* `min_neighbors::Int`: minimum number of neighbors to be a core point
+* `min_cluster_size::Int`: minimum number of points to be a valid cluster
+
+### Output
+* `Vector{DbscanCluster}`: an array of clusters with the id, size core indices and boundary indices
+
+### Example:
+points = randn(3, 10000)
+clusters = dbscan(points, 0.05, min_neighbors = 3, min_cluster_size = 20) # clusters with less than 20 points will be discarded
```
ararslan: Should wrap the example code in triple backticks as in GitHub markdown

REQUIRE
OmriRoames: Yes, PointClouds.jl is actively maintained by FugroRoames. It's a container for large point clouds, and there is no real reason to have it here. My bad.

OmriRoames:
Thanks a lot for the feedback. I've removed the PointClouds dependency

src/dbscan.jl
```diff
@@ -203,19 +199,19 @@ function _update_exploration_list!(adj_list::Array{Int}, exploration_list::Vecto
 end
 
 """
-`_accept_cluster!(clusters, core_selection, cluster_selection)`
+    `accept_cluster!(clusters, core_selection, cluster_selection)`
```
ararslan: When it's indented, it shouldn't still be surrounded in backticks

OmriRoames:
Ok, fixed

ararslan:
Looks good to me.

src/dbscan.jl
```diff
@@ -102,4 +109,109 @@ function _dbs_expand_cluster!{T<:Real}(D::DenseMatrix{T},           # distance m
...
+
+### Example:
+``` julia
+points = randn(3, 10000)
+clusters = dbscan(points, 0.05, min_neighbors = 3, min_cluster_size = 20) # clusters with less than 20 points will be discarded
+```
+"""
+function dbscan{T <: AbstractFloat, N}(points::Array{T, N}, radius::Real; leafsize::Int = 20, kwargs ...)
+    kdtree = KDTree(points; leafsize=leafsize)
+    kdtree::KDTree{Float64, Distances.Euclidean} # TODO: julia-0.4 fails type inference here
```
KristofferC: Should it be a `T` here instead of `Float64`?

OmriRoames: Good point.

JoshChristie:
Should the prior dbscan method be removed/deprecated?
Could you update the docs page with the new function?

OmriRoames:
@ararslan how can I update the documentation?

ararslan:
You can edit doc/source/dbscan.rst

src/Clustering.jl
```diff
@@ -33,7 +34,7 @@ module Clustering
     AffinityPropResult, affinityprop,
 
     # dbscan
-    DbscanResult, dbscan,
+    DbscanResult, DBSCANCluster, dbscan,
```
ararslan: Just noticed: `DBSCANResult` should be capitalized to match `DBSCANCluster`.

ararslan:
@JoshChristie If I'm not mistaken (perhaps Omri can correct me if I am), this change will have no user-facing effects aside from renaming changing the capitalization of types.

OmriRoames:
@ararslan I like DBSCANResult better as well but I am worried that changing it will break peoples code. What do you think of @JoshChristie suggestion to deprecate the old DBSCAN?

ararslan:
Oh my point about non-user facing was that there would be nothing to deprecate. Or have you changed the actual API of the `dbscan` function? For whatever reason it isn't showing any removed code in the diff when I look at it, but there's current a `dbscan` function on master.

OmriRoames:
I haven't removed the existing one. The API for the new function is different: the existing one takes an adjacency matrix (which is what cause the memory to blow up) while the new one simply takes the points themselves (no adjacency matrix involved). I think the existing one should go, but if we remove it it will break peoples code. I would go with deprecation.

ararslan:
Ah okay, I see now. I apologize for my initial confusion.

Deprecating doesn't seem necessary to me; if people want to use the method with adjacency matrices then they can, or they can use your method. I'd be interested to hear what others think about that though. @nalimilan?

JoshChristie:
If both methods are kept it would good to recommend the new method other the old method due to the performance increase for large data.

A quick performance test:
```julia
using Clustering
# Create 3 separate clusters (30,000 points)
a = randn(3,10000)
b = randn(3,10000) .+ 5.0
c = randn(3,10000) .+ 10.0
points = [a b c]

# old method (timings below recorded after a couple of reruns)
using Distances
@time dist_matrix = pairwise(Euclidean(), points, points) 
# 30.865760 seconds (180.01 k allocations: 6.712 GB, 5.62% gc time)
@time dbscan(dist_matrix, 1.0, 1) 
# 2.843919 seconds (380.61 k allocations: 607.347 MB, 2.72% gc time)

# new method
@time dbscan(points, 1.0)
# 1.713170 seconds (501.31 k allocations: 565.157 MB, 1.83% gc time)
````

OmriRoames:
OK, I've changed `DBSCANCluster` back to `DbscanCluster` for consistency, and wrote some documentation.

OmriRoames:
@ararslan is there anything else that needs some attention?

ararslan:
Don't think so. This looks great, thanks for contributing!

PR #73, opened 2016-08-30
remove typeassert in KDTree

KristofferC:
This will no longer be true on 0.5 with new version of `NearestNeighbors`. Since this sits just behind a function barrier the type assert does not actually improves performance and can just be removed.


ararslan:
Looks good to me!

Issue #64, opened 2016-06-08
Nondetermanistic methods should take a n_init parameter, for how many times to run


oxinabox:
Hiya,
In python's sklearn methods like [K-Means](http://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html#sklearn.cluster.KMeans) and K-Mediods take an n_init parameter:
To use their description:

> **n_init : int, default: 10**
>     Number of time the k-means algorithm will be run with different centroid seeds. The final results will be the best output of n_init consecutive runs ...
> 

I think it would be good to have that here,
particularly since a single run of k-means is fairly poor as a method for clustering data.
Running it several times and taking the best is common practice.

ararslan:
Sounds like a good idea to me. Would you be interested in submitting a PR for that?

PR #77, opened 2016-09-30
Issue #64: added n_init to kmeans

lbollar:
Implementing n_init feature from sklearn mentioned in [Issue 64](https://github.com/JuliaStats/Clustering.jl/issues/64).

First ever Julia or package commit, please let me know errors or what needs to be cleaned up.

### DIFF
--- a/src/Clustering.jl
+++ b/src/Clustering.jl
@@ -48,8 +48,10 @@ module Clustering
     randindex,
 
     # hclust
-    Hclust, hclust, cutree
+    Hclust, hclust, cutree,
 
+    # MCL
+    mcl, MCLResult
 
     ## source files
 
@@ -60,6 +62,7 @@ module Clustering
     include("kmedoids.jl")
     include("affprop.jl")
     include("dbscan.jl")
+    include("mcl.jl")
 
     include("silhouette.jl")
     include("randindex.jl")
--- a/src/mcl.jl
+++ b/src/mcl.jl
@@ -1,0 +1,205 @@
+# MCL (Markov CLustering algorithm)
+
+"""
+    immutable MCLResult <: ClusteringResult
+
+Result returned by `mcl()`.
+"""
+immutable MCLResult <: ClusteringResult
+    mcl_adj::Matrix{Float64}    # final MCL adjacency matrix (equailibrium state matrix if converged)
+    assignments::Vector{Int}    # element-to-cluster assignments (n)
+    counts::Vector{Int}         # number of samples assigned to each cluster (k)
+    nunassigned::Int            # number of single elements not assigned to any cluster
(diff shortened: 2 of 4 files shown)


src/kmeans.jl
```diff
@@ -43,18 +45,37 @@ function kmeans(X::Matrix, k::Int;
                 weights=nothing,
                 init=_kmeans_default_init,
                 maxiter::Integer=_kmeans_default_maxiter, 
+				n_init::Integer=_kmeans_default_n_init,
```
ararslan: Please use 4 space indents rather than tabs

src/kmeans.jl
```diff
@@ -17,12 +17,14 @@ const _kmeans_default_init = :kmpp
...
 const _kmeans_default_tol = 1.0e-6
 const _kmeans_default_display = :none
+const _kmeans_default_n_init = 10
 
 function kmeans!{T<:AbstractFloat}(X::Matrix{T}, centers::Matrix{T};
                                    weights=nothing,
                                    maxiter::Integer=_kmeans_default_maxiter, 
                                    tol::Real=_kmeans_default_tol,
                                    display::Symbol=_kmeans_default_display)
+
```
ararslan: Unrelated (and unnecessary) change--best to keep PRs focused on the task at hand

src/kmeans.jl
```diff
@@ -17,12 +17,14 @@ const _kmeans_default_init = :kmpp
 const _kmeans_default_maxiter = 100
 const _kmeans_default_tol = 1.0e-6
 const _kmeans_default_display = :none
+const _kmeans_default_n_init = 10
```
ararslan: Out of curiosity, what's the motivation for a default of 10?

src/kmeans.jl
```diff
@@ -43,18 +45,37 @@ function kmeans(X::Matrix, k::Int;
...
     m, n = size(X)
     (2 <= k < n) || error("k must have 2 <= k < n.")
-    iseeds = initseeds(init, X, k)
-    centers = copyseeds(X, iseeds)
-    kmeans!(X, centers; 
-            weights=weights, 
-            maxiter=maxiter,
-            tol=tol,
-            display=display)
+	n_init > 0 || error("n_init must be greater than 0")
```
ararslan: Since `n_init` is a function argument, the error should be an `ArgumentError`. You can change this to
```julia
n_init > 0 || throw(ArgumentError("n_init must be greater than 0"))
```

src/kmeans.jl
lbollar: 10 was chosen as this is the default for sklearn n_init param, and since we are duplicating this functionality, I just took this value.

PR #82, opened 2017-02-27
fix 0.6 deprecations + drop 0.4

CarloLucibello:


REQUIRE
```diff
@@ -1,5 +1,5 @@
-julia 0.4
-Compat 0.8.5
-Distances 0.3.1
-NearestNeighbors 0.0.3
-StatsBase 0.9.0
```
ararslan: The lower bounds on the dependencies should not be removed—those are important to have there

CarloLucibello: right. I put them back, and raised the lower bound for Compat

CarloLucibello:
no idea of how to fix the breaking on osx nightly

CarloLucibello:
I've seen this macos + 0.6 failure in other repos, it is an upstream problem. Tests are passing on other machines. Merge?

ararslan:
I don't think I've seen a precision issue specific to Julia 0.6 on macOS. Can you point me to somewhere else you've seen it? In the meantime, cc @simonbyrne, the numerical accuracy guru.

CarloLucibello:
right, I saw a problem in julia 0.6 on macos here https://github.com/JuliaGraphics/Gtk.jl/pull/287, but that is something complitely different, related to homebrew. I've no idea of what's going on here

src/dbscan.jl
```diff
@@ -208,9 +208,11 @@ Accept cluster and update the clusters list
 * `core_selection :: Vector{Bool}`: selection of the core points of the cluster
 * `cluster_selection :: Vector{Bool}`: selection of all the cluster points
 """
-function accept_cluster!(clusters::Vector{DbscanCluster}, core_selection::BitArray{1}, cluster_selection::BitArray{1}, cluster_size::Int)
+function accept_cluster!(clusters::Vector{DbscanCluster}, core_selection::BitVector,
+        cluster_selection::BitVector, cluster_size::Int)
```
nalimilan: Indentation is incorrect here and below.

src/kmeans.jl
```diff
@@ -153,8 +153,8 @@ function _kmeans!{T<:AbstractFloat}(
         end
     end
 
-    return KmeansResult(centers, assignments, costs, counts, cweights, 
-            @compat(Float64(objv)), t, converged)
+    return KmeansResult(centers, assignments, costs, counts, cweights,
+                    Float64(objv), t, converged)
```
nalimilan: Here too.

src/mcl.jl
```diff
@@ -92,7 +92,7 @@ function _mcl_inflate!(dest::Matrix{Float64}, src::Matrix{Complex128}, inflation
     src_norm = vecnorm(src)
     min_rel = -1E-3*src_norm
     min_img = 1E-3*src_norm
-    @inbounds for (i, el) in enumerate(src)
+    for (i, el) in enumerate(src)
```
nalimilan: Keep `@inbounds` here, it makes a difference (even if ideally the fact that `enumerate` is used would allow skipping the checks). In general, no unrelated changes should be made in a PR.

CarloLucibello: ouch, this was unintended

src/dbscan.jl
CarloLucibello: below where?

test/kmedoids.jl
```diff
@@ -30,14 +30,14 @@ R = kmedoids(costs, k)
...
 costs = pairwise(SqEuclidean(), X)
 
 R = kmedoids!(costs, [1, 2, 6])
 @test isa(R, KmedoidsResult)
 @test R.medoids == [3, 5, 6]
 @test R.assignments == [1, 2, 1, 1, 2, 3, 2, 3, 3]
 @test R.counts == [3, 3, 3]
-@test_approx_eq R.acosts [1, 1, 0, 1, 0, 0, 1, 1, 1]
-@test_approx_eq R.totalcost 6.0
+@test all(isapprox.(R.acosts, [1, 1, 0, 1, 0, 0, 1, 1, 1]))
```
ararslan: Are these dot broadcasts in the tests the only things keeping this from also working on 0.4?

nalimilan: All the `@compat(Union{..})` are also not supported on 0.4. At this point, I wouldn't care about 0.4. And it's not like the new Clustering.jl added many new features 0.4 users may want to use. :-)

ararslan: No, `@compat(Union{...})` is for 0.3 support; the 0.3-style `Union(...)` is deprecated in 0.4.

I'm not opposed to dropping 0.4 support, I was mostly just wondering whether it was actually necessary to allow 0.6 to work.

nalimilan: Ah, good catch!

REQUIRE
```diff
@@ -1,5 +1,5 @@
-julia 0.4
-Compat 0.8.5
+julia 0.5
+Compat 0.18
```
tkelman: shouldn't this be 0.17? don't see any parametric aliases

PR #85, opened 2017-04-16
Allow for arbitrary distance metrics for K-means clustering

jmsteitz:
Adds keyword argument 'distance' to kmeans() and kmeans!() that defaults to Squared Euclidean but allows to specify another semi-metric to use for clustering. Additionally the user needs to provide a custom evaluate(dist, a, b) function for metrics that are not already available in Distances.jl.

### DIFF
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -22,7 +22,8 @@ function kmeans!{T<:AbstractFloat}(X::Matrix{T}, centers::Matrix{T};
                                    weights=nothing,
                                    maxiter::Integer=_kmeans_default_maxiter,
                                    tol::Real=_kmeans_default_tol,
-                                   display::Symbol=_kmeans_default_display)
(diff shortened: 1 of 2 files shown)


ararslan:
Cool, I like it! Thanks for the contribution! It would be great if you could add a test that verifies the correct behavior when passed a different distance metric.

jmsteitz:
Thanks for your positive feedback! I've added tests which pass a copy of standard SqEuclidean distance as a custom type to be able to compare the expected outcome.

Also I figured out that it's probably easier to just provide a custom pairwise! function instead of evaluate. Additionally in the case of SqEuclidean distance the outcome differs, because the structure of pairwise! for the SqEuclidean distance is different from the generic one in Distances.jl.

ararslan:
Looks good to me. I'll leave this open for a bit to give other maintainers the chance to comment, but as far as I'm concerned it's good to go.

PR #89, opened 2017-08-17
Added fuzzy c-means

annimesh2809:
An implementation of the standard fuzzy c-means algorithm as stated [here](https://github.com/JuliaImages/ImageSegmentation.jl/pull/11). I have used [this](http://yaikhom.com/res/fcm/fcm.pdf) paper as a reference.


ararslan:
Yes, for now I'd say you should make them 0.5 compatible. We can later drop 0.5 support and have FemtoCleaner handle the 0.5 -> 0.6 syntax changes.

kmsquire:
@annimesh2809, looks good.  Could you add some tests?

kmsquire:
@annimesh2809, thanks for adding tests. 

The tests are failing on v0.5 because `.=` isn't defined there, and on master because of a static compilation problem with StaticArrays. V0.6 isn't being tested by Travis. 

@ararslan, you seem pretty busy, but do you think you might be able to update Travis testing here?  I might be able to get to it this weekend, but not sure. 

After that's fixed, my opinion is that this package (and pull request) could be upgraded to only target v0.6 and later. It doesn't have any very active maintainers, and that just makes things easier. 

@annimesh2809, if you wanted it merged sooner, you could change all uses of `.=` to simply be `=` for now. Performance will probably go down for v0.6 and v0.7, but testing would probably pass. Or you can wait until this package is bumped to v0.6. You're also very welcome to bump the package to support only v0.6 and later yourself (in a separate pull request).

annimesh2809:
@kmsquire I have updated the tests and now they pass successfully in Julia v0.5. I think it would be better to merge this first (as some more [PR's](https://github.com/JuliaImages/ImageSegmentation.jl/pull/11) depend on this now.) and later I will bump this package to v0.6 in a separate PR.

Also could you kindly take a look at issue #91. If you want, I can add the fix (i.e. change `kmeans(X::Matrix, ...)` to `kmeans{T<:AbstractFloat}(X::Matrix{T}, ...)` ) in this PR itself.

kmsquire:
@annimesh2809, thank you. I hadn't realized that the v0.5 error was in the tests themselves. I'll merge this when I get to a computer (unless someone beats me to it). 

Regarding issue #91, can you please open up another pull request with a fix?  We ideally like to keep unrelated changes in different commits (at least) and generally different pull requests.

src/fuzzycmeans.jl
```diff
@@ -0,0 +1,112 @@
+# Fuzzy C means algorithm
+
+## Interface
+
+immutable FuzzyCMeansResult{T<:AbstractFloat} <: ClusteringResult
+    centers::Matrix{T}          # cluster centers (d x C)
+    weights::Matrix{T}          # assigned weights (n x C)
```
kmsquire: If you make `centers` take on the type of the input `data`, then it's possible that `weights` will have a different type (e.g., in k-means, the centers have a parameterized type `T`, but the weight vector is a `Float64`).  How about doing the same thing here?

src/fuzzycmeans.jl
```diff
@@ -0,0 +1,112 @@
...
+            end
+            weights[i,j] = 1.0/den
+        end
+    end
+end
+
+function update_centers!(centers, data, weights, fuzziness)
+    nrows, ncols = size(weights)
+    for j in 1:ncols
+        num = zeros(Float64, size(data[:,1]))
```
kmsquire: `data[:,1]` makes a copy here.  Use `size(data, 1)`.

Also, if you update `centers` below, then this should use `zeros(T, size(data,1))` (after getting the eltype of centers, of course).

src/fuzzycmeans.jl
```diff
@@ -0,0 +1,112 @@
...
+        den = 0.0
```
kmsquire: `den = zero(T)`

(Not actually sure this is still necessary--can someone say for sure?

src/fuzzycmeans.jl
```diff
@@ -0,0 +1,112 @@
...
+    weights ./= sum(weights, 2)
+
+    centers = fill(0.0, nrows, C)
+    prev_centers = identity.(centers)
+
+    δ = Inf
+    iter = 0
+
+    if displevel >= 2
+        prog = ProgressThresh(tol, "Converging:")
```
kmsquire: How slow is the algorithm here?  While I imagine `ProgressMeter` produces nice output , in most core packages like this, there is often a goal not to pull in too many external dependencies.  Is this output essential, or can you produce something reasonable without `ProgressMeter.jl`?

annimesh2809:
@kmsquire I have replaced `ProgressMeter` with simple `printf` statements similar to `kmeans`. I have also made some other necessary changes.

kmsquire:
@annimesh2809, could you add some documentation for this under docs (in a new pull request)?

annimesh2809:
Yes, sure! But it might take some time because my midterms are coming. WIll that be okay?

ararslan:
Yep, that's fine. Good luck on your midterms!

timholy:
Would it be possible to get a new release?

ararslan:
[The deed is done](https://github.com/JuliaLang/METADATA.jl/pull/10940).

PR #60, opened 2016-02-22
Document Hierarchical Clustering

ahwillia:
I noticed that the `hclust()` function was missing from the main documentation. This is a rough pass at it.

One problem I have -- the Hierarchical Clustering page doesn't show up in the left sidebar. Any advice on how to fix this? It is included in the TOC on the `algorithms.rst` and is visible on that page.

Fixes https://github.com/JuliaStats/Clustering.jl/issues/95.

PR #94, opened 2017-10-16
Ward hierarchical clustering

jmboehm:
I added a routine for Ward hierarchical clustering to hclust.jl, along with tests. The algorithm is from Murtagh, F. and P. Legendre, “Ward’s hierarchical agglomerative clustering method: which algorithms implement Ward’s criterion?” J. Classification 31(3) pp. 274--295, 2014. This is also the base of the code in the R function hclust.

Apologies if something is missing or not well done. I'm still somewhat unfamiliar with contributing code (in fact this is the first time that I'm submitting a pull request).


src/hclust.jl
```diff
@@ -334,6 +467,7 @@ function hclust{T<:Real}(d::Symmetric{T}, method::Symbol)
     inds = Any[]
     merge = h[1]
     for i in 1:size(merge)[1]
+
```
nalimilan: ?

src/hclust.jl
```diff
@@ -316,6 +316,137 @@ function hclust2{T<:Real}(d::Symmetric{T}, method::Function)
     hcat(mr[o], mc[o]), h[o]
 end
 
+# Ward linkage algorithm, using a Lance-Williams formula.
+#
+# Based on the R-coded version of the Fortran function hclust.f, originally
```
nalimilan: What is the license of that file? We can only include code which is licensed under MIT or a compatible license.

src/hclust.jl
```diff
@@ -316,6 +316,137 @@ function hclust2{T<:Real}(d::Symmetric{T}, method::Function)
...
+    h = Vector{T}(nc-1)                  # height
+
+    triindex(n::Int64, i::Int64,j::Int64) = j + (i-1)*n - ((i*(i+1)) >> 1);
+
+    if method==:ward1
+        dscpy = [ds[r,c] for r=1:(nc-1) for c=(r+1):nc]
+    elseif method == :ward2
+        dscpy = [ds[r,c] for r=1:(nc-1) for c=(r+1):nc].^2
+    else
+        error("method ", method, " is unsupported by hclust_lw().")
```
nalimilan: Throw `ArgumentError`. No need to repeat the name of the function, but list supported options.

src/hclust.jl
```diff
@@ -316,6 +316,137 @@ function hclust2{T<:Real}(d::Symmetric{T}, method::Function)
...
+    mr = Vector{Int64}(nc-1)               # min row
+    mc = Vector{Int64}(nc-1)               # min col
+    h = Vector{T}(nc-1)                  # height
+
+    triindex(n::Int64, i::Int64,j::Int64) = j + (i-1)*n - ((i*(i+1)) >> 1);
+
+    if method==:ward1
+        dscpy = [ds[r,c] for r=1:(nc-1) for c=(r+1):nc]
+    elseif method == :ward2
+        dscpy = [ds[r,c] for r=1:(nc-1) for c=(r+1):nc].^2
```
nalimilan: Better square inside the comprehension to avoid making a copy.

src/hclust.jl
```diff
@@ -316,6 +316,137 @@ function hclust2{T<:Real}(d::Symmetric{T}, method::Function)
...
+    else
+        error("method ", method, " is unsupported by hclust_lw().")
+    end
+
+    merges = -collect(1:nc)
```
nalimilan: Put the `-` inside the call to avoid making a copy.

src/hclust.jl
```diff
@@ -316,6 +316,137 @@ function hclust2{T<:Real}(d::Symmetric{T}, method::Function)
...
+    membr = ones(Float64, nc)
```
nalimilan: `ones` and `zeros` default to `Float64`, no need to specific it (here and elsewhere).

src/hclust.jl
```diff
@@ -316,6 +316,137 @@ function hclust2{T<:Real}(d::Symmetric{T}, method::Function)
...
+    checkme = trues(nc)
+    nearestneighbor = zeros(Int64, nc-1)
+    distancetonearestneighbor = zeros(Float64, nc-1)
```
nalimilan: Could you find shorter names? :-)

src/hclust.jl
```diff
@@ -316,6 +316,137 @@ function hclust2{T<:Real}(d::Symmetric{T}, method::Function)
...
+        distancetonearestneighbor[i] = dmin
+    end
+
+    # main clustering loop
+    for nclust = nc:-1:2
+
+        # check list of nearest neighbors to determine next merge
+        dmin = Inf
+        for i=1:(nc-1)
+            if (checkme[i]==true)
```
nalimilan: As a matter of style, we don't use parentheses in `if` (same elsewhere). But add spaces around `==`, `!=` and `=`, and after commas.

src/hclust.jl
```diff
@@ -316,6 +316,137 @@ function hclust2{T<:Real}(d::Symmetric{T}, method::Function)
...
+                    ind1 = triindex(nc,i2,k)
+                else
+                    ind1 = triindex(nc,k,i2)
+                end
+                if (j2<k)
+                    ind2 = triindex(nc,j2,k)
+                else
+                    ind2 = triindex(nc,k,j2)
+                end
+                dscpy[ind1] = ((membr[i2] + membr[k])*dscpy[ind1] + (membr[j2] + membr[k])*dscpy[ind2] - membr[k]*dscpy[ind3]) / (membr[i2]+membr[j2]+membr[k])
```
nalimilan: Avoid lines above 92 characters.

src/hclust.jl
```diff
@@ -316,6 +316,137 @@ function hclust2{T<:Real}(d::Symmetric{T}, method::Function)
...
+                nearestneighbor[i] = jj
+                distancetonearestneighbor[i] = dmin
+            end
+        end
+
+    end # main clustering loop
+
+    # take root if we use ward2
+    if (method == :ward2)
+        h=sqrt.(h)
```
nalimilan: Use `h .= sqrt.(h)` to avoid making a copy (and possibly introducing a type instability).

src/hclust.jl
```diff
@@ -326,6 +457,8 @@ function hclust{T<:Real}(d::Symmetric{T}, method::Symbol)
         h = hclust2(d, slicemaximum)
     elseif method == :average
         h = hclust2(d, slicemean)
+    elseif (method == :ward1) || (method == :ward2)
```
nalimilan: Have you thought about the best names? IIUC, `ward1` was the only option in R for a long time, but it's said not to implement the algorithm actually proposed by Ward. So should it be called `ward.wrong` (joking, but you see the point)? Is there any reason to prefer one algorithm over the other one? R package agnes uses `ward2` by default, and calls it `ward`. We should probably check what other software does.

nalimilan:
Ah, before writing the docs, please tell me whether https://github.com/JuliaStats/Clustering.jl/pull/60 is still up-to-date, and I can merge it so that you base your work on it.

src/hclust.jl
jmboehm: hclust.f is distributed with R:
https://github.com/wch/r-source/blob/trunk/src/library/stats/src/hclust.f
While there is no licence described on that file itself, the "surrounding" code is under the GPL. Do you think that this is a problem?

Thanks for your detailed comments on the code.

src/hclust.jl
jmboehm: According to https://link.springer.com/article/10.1007/s00357-014-9161-z Statistica, Systat, and SPSS use ward1, whereas Matlab, SAS, and JMP use ward2. R have changed their names to "ward.D" and "ward.D2". The above paper uses "Ward1" and "Ward2", so I thought it was reasonable to stick with that.

src/hclust.jl
nalimilan: Argh. Yes, that's a showstopper for inclusion into this package, as work derived from GPL code is itself licensed under the GPL, which is not compatible with the MIT license. So there are three options:
1) Relicense the package to GPL (other files can remain MIT, but the full package must be GPL). I'm afraid it's original authors wouldn't be fans of this solution.
2) Ship the code in a separate package. This of course has the drawback that it would be harder to discover. Note that  @davidavdav's HClust.jl, which has been merged into this package, was GPL, but I guess he didn't take inspiration from the R code. You could still revive that package and add your function there.
3) Get the authorization from the authors to relicense the code as MIT. Since that's not directly their code and Julia is open source, they might be open to this solution. You could contact Martin Maechler, I think he's discussed with Julia developers in the recent years. The two only other listed authors are Fritz Leisch and Ross Ihaka, so they could be contacted without too much work too.

If none of these possibilities works, somebody will have to rewrite the code only based on descriptions of the algorithms which are publically available. You could describe in detail what the R code does, as long as the person who write the code doesn't copy it directly. Yet's, that's really painful...

src/hclust.jl
nalimilan: OK. But since Ward 2 is the actual method proposed by Ward, probably better use it by default? I wonder whether there are possible implications when it's used in combination with PCA, which is also based on variance.

src/hclust.jl
davidavdav: For my original `HClust.jl`, I did not look at any code, just at the scientific articles quoted in the code.  I changed the license implicitly with the PR that included the code in this repository.

jmboehm: Thanks. I should have really thought about these licensing issues before...
Is it generally sufficient to get permissions from one of the authors of the original code (say, the package maintainer), or do I need to get authorization from all of them?

nalimilan: From all of them... :-/ At least we're lucky, the number of contributors is quite small (compare with e.g. this repo).

davidavdav: I've had experience before asking for a change of licence from an original author with `MFCC.jl`, and that was quite easy.  But that was just a single author.

nalimilan: @jmboehm Have you tried contacting the authors in the end?

jmboehm: Yes. The original author, Fionn Murtagh, seemed initially very positive about it, but had some questions about the MIT license. When I explained it to him, he stopped responding. So I guess that means no. I'll fork my PR into a separate (GPL'ed) repo.

jmboehm:
Closing because of licensing issues.

src/hclust.jl
nalimilan: Probably worth asking once again? We're all very busy, and it's easy to forget to reply to a message when you have no direct interest in it. You could also try to contact other authors, as it's a convincing argument to say "your coauthors have already accepted".

jmboehm: I'll ask again, and will re-open if I get a positive answer.

nalimilan:
The Ward method has been implemented by https://github.com/JuliaStats/Clustering.jl/pull/106.

PR #96, opened 2017-10-17
Cosmetic changes in hclust.jl

davidavdav:
Just some spacing, `≤` rather than `<=`, and some double-for-loop rewrites in my own code.

### DIFF
--- a/test/hclust.jl
+++ b/test/hclust.jl
@@ -1,3 +1,5 @@
+using Clustering
+using Base.Test
 
 # load the examples array
 include("hclust-generated-examples.jl")
--- a/src/hclust.jl
+++ b/src/hclust.jl
@@ -1,4 +1,4 @@
-## hclust.jl (c) 2014 David A. van Leeuwen
+## hclust.jl (c) 2014, 2017 David A. van Leeuwen
 ## Hierarchical clustering, similar to R's hclust()
 
 ## Algorithms are based upon C. F. Olson, Parallel Computing 21 (1995) 1313--1325.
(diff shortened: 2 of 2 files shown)


test/hclust.jl
```diff
@@ -1,3 +1,5 @@
+using Clustering
```
nalimilan: If that was needed I guess the tests would fail?

nalimilan:
+1 I prefer `in` too!

ararslan:
Heh, I actually like `=` more than `in` but I don't care that much. 😉 I still appreciate the contribution!

test/hclust.jl
davidavdav: The test as run from `src/runtests.jl` would work all right, but this line allows a separate test
```julia
include("src/Clustering.jl")
include("test/hclust.jl")
```

ararslan: I assume this isn't the cause of the nightly failure?
```
ERROR: LoadError: LoadError: importing Clustering into Main conflicts with an existing identifier
```

davidavdav: All the other tests in `test/` have `using Clustering` in their code (but after `using Base.Test`---I don't assume order can matter here).  

On my machine, it is `mcl.jl` that fails on an `InexactError()`.  But that is a different issue I suppose.

ararslan: Huh, weird. Okay.

nalimilan: I'd check whether the error hasn't been introduced by this PR though, and if so file an issue in Base.

alyst:
I guess the only thing that keeps the PR from merging are Travis failures.
@davidavdav Could you please rebase it to the current master to check whether the tests pass on 0.6?

davidavdav:
I have tried, but git / hub /travis have defied me.

alyst:
Thanks! The tests pass for 0.6 on Linux, so I'm merging.

PR #98, opened 2017-10-23
Update kmedoids.jl

nan2ge1:
Changed field name of KmedoidsResult to mirror that of KmeansResult.


wildart:
@nangeschnappt Corresponding tests require fix as well.

PR #99, opened 2017-11-17
bug fix in randindex

trappmartin:
RandIndex should be 1. of both assignments are equal. The current code returns 0.

### DIFF
--- a/src/randindex.jl
+++ b/src/randindex.jl
@@ -26,7 +26,11 @@ function randindex(c1,c2)
 
     if t1 == nc
         # avoid division by zero; if k=1, define Rand = 0
-        ARI = 0
+        if n == 1
+            ARI = 0.
+        else
+            ARI = 1.
+        end
     else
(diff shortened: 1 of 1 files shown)


kmsquire:
Hi @trappmartin, thanks for the contribution.  Can you give an example of where the current code is incorrect and your code is correct?  I'm having trouble creating one.  

If I have, e.g.

```julia
c1 = [1]
c2 = [1]
```

then `n == 1`, but `nc == nan`, so `t1 != nc`, and the test on line 27 fails, and produces a `nan` result... so it seems that the code is still incorrect?

kmsquire:
Also, could you also add tests which cover all branches of your changes (once you've verified that they're correct)?

trappmartin:
Interesting, I haven't checked this case.
On my machine, the current code produces `0` if
```julia
c1 = [1, 2, 3]
c2 = [1, 2, 3]
```
which should not be the case. 
I'll check the case you reported and add a fix asap (after the weekend).

PR #100, opened 2017-11-23
Update travis config

wildart:
Update travis config to support 0.6 and allow failures for nightly builds


ararslan:
Sure, thanks. Though at some point we should drop 0.5 and fix whatever is broken on 0.7, given that feature freeze is soon.

PR #101, opened 2017-11-23
Fix type inconsistency in dbscan call

### LINKED ISSUE #84: DBSCAN gives error on Float32 array
I can only use the DBSCAN algorithm with a Float64 array but not with a Float32 array, that is actually my input.

`
using Clustering; positions = zeros(Float32, 3, 10); clusters = dbscan(positions, 0.3, min_neighbors=1, min_cluster_size=1, leafsize=20)
`

ERROR: MethodError: no method matching _dbscan(::NearestNeighbors.KDTree{StaticArrays.SVector{3,Float32},Distances.Euclidean,Float32}, ::Array{Float32,2}, ::Float64; min_neighbors=1, min_cluster_size=1)
Closest candidates are:
  _dbscan{T<:AbstractFloat}(::NearestNeighbors.KDTree{V<:AbstractArray{T,1},M<:Union{Distances.Chebyshev,Distances.Cityblock,Distances.Euclidean,Distances.Minkowski},T}, ::Array{T<:AbstractFloat,2}, ::T<:AbstractFloat; min_neighbors, min_cluster_size) at /user/.julia/v0.5/Clustering/src/dbscan.jl:144
 in #dbscan#6(::Int64, ::Array{Any,1}, ::Function, ::Array{Float32,2}, ::Float64) at /user/.julia/v0.5/Clustering/src/dbscan.jl:137
 in (::Clustering.#kw##dbscan)(::Array{Any,1}, ::Clustering.#dbscan, ::Array{Float32,2}, ::Float64) at ./<missing>:0

wildart:
This PR fixes type inconsistency in dbscan (fixes #84)

### DIFF
--- a/src/dbscan.jl
+++ b/src/dbscan.jl
@@ -132,15 +132,15 @@ points = randn(3, 10000)
 clusters = dbscan(points, 0.05, min_neighbors = 3, min_cluster_size = 20) # clusters with less than 20 points will be discarded
 ```
 """
-function dbscan{T <: Real}(points::AbstractMatrix{T}, radius::T; leafsize::Int = 20, kwargs ...)
+function dbscan(points::AbstractMatrix, radius::Real; leafsize::Int = 20, kwargs ...)
     kdtree = KDTree(points; leafsize=leafsize)
     return _dbscan(kdtree, points, radius; kwargs ...)
 end
 
 
 """ An implementation of DBSCAN algorithm that keeps track of both the core and boundary points """
-function _dbscan{T<:Real}(kdtree::KDTree, points::Matrix{T}, radius::T;
-                          min_neighbors::Int = 1, min_cluster_size::Int = 1)
+function _dbscan(kdtree::KDTree, points::AbstractMatrix, radius::Real;
+                 min_neighbors::Int = 1, min_cluster_size::Int = 1)
     dim, num_points = size(points)
     num_points <= dim && throw(ArgumentError("points has $dim rows and $num_points columns. Must be a D x N matric with D < N"))
     0 <= radius || throw(ArgumentError("radius $radius must be ≥ 0"))
--- a/src/dbscan.jl
+++ b/src/dbscan.jl
@@ -132,15 +132,15 @@ points = randn(3, 10000)
 clusters = dbscan(points, 0.05, min_neighbors = 3, min_cluster_size = 20) # clusters with less than 20 points will be discarded
 ```
 """
-function dbscan{T <: AbstractFloat, N}(points::Array{T, N}, radius::Real; leafsize::Int = 20, kwargs ...)
+function dbscan{T <: Real}(points::AbstractMatrix{T}, radius::T; leafsize::Int = 20, kwargs ...)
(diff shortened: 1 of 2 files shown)


src/dbscan.jl
```diff
@@ -132,15 +132,15 @@ points = randn(3, 10000)
 clusters = dbscan(points, 0.05, min_neighbors = 3, min_cluster_size = 20) # clusters with less than 20 points will be discarded
 ```
 """
-function dbscan{T <: AbstractFloat, N}(points::Array{T, N}, radius::Real; leafsize::Int = 20, kwargs ...)
+function dbscan{T <: Real}(points::AbstractMatrix{T}, radius::T; leafsize::Int = 20, kwargs ...)
```
alyst: I think `radius::T` is too restrictive, and it would be nice to support different types for `points` and `radius`.
In fact, it's also too restrictive in `_dbscan()` (which is the cause of #84).
There, `radius` is compared with the result of `NearestNeigbours.inrange()`, which might have different type than `T`.
So I'd rather set it to `::Real` in both methods.

src/dbscan.jl
```diff
@@ -132,15 +132,15 @@ points = randn(3, 10000)
...
     kdtree = KDTree(points; leafsize=leafsize)
     return _dbscan(kdtree, points, radius; kwargs ...)
 end
 
 
 """ An implementation of DBSCAN algorithm that keeps track of both the core and boundary points """
-function _dbscan{T<:AbstractFloat}(kdtree::KDTree, points::Matrix{T}, radius::T;
-                                   min_neighbors::Int = 1, min_cluster_size::Int = 1)
+function _dbscan{T<:Real}(kdtree::KDTree, points::Matrix{T}, radius::T;
```
alyst: `radius::Real`

We can also use this opportunity and improve the signature:
* remove `T` completely (it's not used in the body)
* set `points` to `::AbstractMatrix` to match `dbscan()`
* it might be also good to make `min_neighbors` and `min_cluster_size` positional since this method is internal and avoiding keyargs should improve type inference, but that's out of the scope of fixing #84

test/dbscan.jl
```diff
@@ -46,3 +46,9 @@ clusters = dbscan(points, 0.3, min_neighbors=2, min_cluster_size=100, leafsize=2
 @test clusters[1].core_indices == inds_1
 @test clusters[2].core_indices == inds_2
 @test clusters[3].core_indices == inds_3
+
+# Issue #84
+clusters = dbscan(convert(Matrix{Float32}, points), 0.3f0, min_neighbors=2, min_cluster_size=100, leafsize=20)
```
alyst: Since we are allowing any radius, it would be nice to test that `Matrix{Float32}` / `Float64` combination also works (the actual error in #84).

src/dbscan.jl
wildart: I believe `Real` for radius is quite sufficient, it's a positive real number. For points, the parameterization can be drop.

test/dbscan.jl
wildart: I added one.

alyst:
Thanks!

PR #103, opened 2017-11-26
Remove inexact exception in favor of values pruning in MCL

### LINKED ISSUE #87: test failure on OS X with 0.6
The cause is `mcl`: here is a simple reproducible example:
```
adj_matrix = [1.0 0.125 1.0 0.0 0.16 0.0; 0.125 1.0 0.0 0.25 0.0 0.16; 1.0 0.0 1.0 0.0 0.2 0.0; 0.0 0.25 0.0 1.0 0.0 0.5; 0.16 0.0 0.2 0.0 1.0 0.0; 0.0 0.16 0.0 0.5 0.0 1.0]
mcl(adj_matrix, display=:verbose, inflation=1.5, expansion=1.5, save_final_matrix=true)
```

I had a quick look through, my guess is that there is a slight change in BLAS behaviour which causes `eig` to give subtly different values, and that this error then amplifies, but i don't understand the algorithm well enough to be sure.

wildart:
Currently, if adjacency matrix values become small, during the inflation step, an inexact exception is thrown and algorithm is terminated. This PR removes this inexact exception, and substitutes it with a threshold pruning schema for column values which guarantees convergence of the algorithm.

This will fix #87 and recent CI test failures.

Ref: see 11.1.1 in "Graph clustering by flow simulation" by Stijn van Dongen.

### DIFF
--- a/src/mcl.jl
+++ b/src/mcl.jl
@@ -5,8 +5,8 @@
 
 Result returned by `mcl()`.
 """
-immutable MCLResult <: ClusteringResult
-    mcl_adj::Matrix{Float64}    # final MCL adjacency matrix (equilibrium state matrix if converged)
+immutable MCLResult{T<:Real} <: ClusteringResult
+    mcl_adj::AbstractMatrix{T}    # final MCL adjacency matrix (equilibrium state matrix if converged)
     assignments::Vector{Int}    # element-to-cluster assignments (n)
     counts::Vector{Int}         # number of samples assigned to each cluster (k)
     nunassigned::Int            # number of single elements not assigned to any cluster
@@ -19,7 +19,7 @@ end
 # Return the tuple: cluster indices for each element, cluster sizes,
 # the number of unassigned (0 cluster index) elements (if `allow_singles` is on)
 # `zero_tol` is a minimal value to consider as an element-to-cluster assignment
-function _mcl_clusters(mcl_adj::Matrix{Float64}, allow_singles::Bool, zero_tol::Float64 = 1E-20)
+function _mcl_clusters(mcl_adj::AbstractMatrix, allow_singles::Bool, zero_tol::Float64 = 1E-20)
     # remove rows containing only zero elements and convert into a mask of nonzero elements
     el2clu_mask = mcl_adj[squeeze(sum(mcl_adj, 2), 2) .> zero_tol, :] .> zero_tol
 
@@ -69,23 +69,43 @@ function _mcl_clusters(mcl_adj::Matrix{Float64}, allow_singles::Bool, zero_tol::
     clu_ixs, clu_sizes, unassigned_count
 end
 
-# adjacency matrix inflation (element-wise raising to a given power) kernel
-function _mcl_inflate!{T<:Real}(dest::Matrix{Float64}, src::Matrix{T}, inflation::Integer)
-    map!(el -> el^inflation, dest, src)
+# adjacency matrix expansion (matrix-wise raising to a given power) kernel
+function _mcl_expand(src::AbstractMatrix, expansion::Number)
+    try
+        return src^expansion
+    catch ex # FIXME: remove this when functionality become available in the standard library
+        if isa(ex, MethodError)
+            throw(ArgumentError("MCL expansion of $(typeof(src)) with expansion=$expansion not supported"))
+        else
+            rethrow()
+        end
+    end
 end
 
+# integral power inflation
+el_inflate(el::Number, inflation::Integer) = el^inflation
+
+# non-integral power inflation
+el_inflate(el::Number, inflation::Number) = real((el+0im)^inflation)
+
 # adjacency matrix inflation (element-wise raising to a given power) kernel
-function _mcl_inflate!{T<:Number}(dest::Matrix{Float64}, src::Matrix{T}, inflation::Number)
-    map!(el -> real((el+0im)^inflation), dest, src)
+function _mcl_inflate!(dest::AbstractMatrix, src::AbstractMatrix, inflation::Number)
+    @inbounds for i in eachindex(src)
+        dest[i] = el_inflate(src[i], inflation)
+    end
 end
 
 # adjacency matrix pruning
-function _mcl_prune!(src::Matrix{Float64}, prune_tol::Number)
-    for i in 1:size(src,2)
-        c = view(src, :, i)
+function _mcl_prune!{T<:Number}(mtx::AbstractMatrix{T}, prune_tol::Number)
+    for i in 1:size(mtx,2)
+        c = view(mtx, :, i)
         θ = mean(c)*prune_tol
-        c[c .< θ] = 0.0
+        @inbounds @simd for j in eachindex(c)
+            c[j] = ifelse(c[j] >= θ, c[j], 0.0)
+        end
     end
+    issparse(mtx) && dropzeros!(mtx)
+    return mtx
 end
 
 """
@@ -105,18 +125,20 @@ Identify clusters in the weighted graph using Markov Clustering Algorithm (MCL).
   MCL doesn't converge
 * `max_iter::Integer`: max number of MCL iterations
 * `tol::Number`: MCL adjacency matrix convergence threshold
-# `prune_tol::Number`: Threshold pruning constant
+* `prune_tol::Number`: pruning threshold
 * `display::Symbol`: `:none` for no output or `:verbose` for diagnostic messages
 
 See [original MCL implementation](http://micans.org/mcl).
+
+Ref: Stijn van Dongen, "Graph clustering by flow simulation", 2001
 """
-function mcl{T<:Number}(adj::Matrix{T};
-                        add_loops::Bool = true,
-                        expansion::Number = 2, inflation::Number = 2.0,
-                        save_final_matrix::Bool = false,
-                        allow_singles::Bool = true,
-                        max_iter::Integer = 100, tol::Number=1.0e-5,
-                        prune_tol::Number=1.0e-5, display::Symbol=:none)
+function mcl{T<:Real}(adj::AbstractMatrix{T};
+                      add_loops::Bool = true,
+                      expansion::Number = 2, inflation::Number = 2.0,
+                      save_final_matrix::Bool = false,
+                      allow_singles::Bool = true,
+                      max_iter::Integer = 100, tol::Number=1.0e-5,
+                      prune_tol::Number=1.0e-5, display::Symbol=:none)
     m, n = size(adj)
     m == n || throw(DimensionMismatch("Square adjacency matrix expected"))
 
@@ -144,7 +166,7 @@ function mcl{T<:Number}(adj::Matrix{T};
     converged = false
     rel_delta = NaN
     while !converged && niter < max_iter
-        expanded = mcl_adj^expansion
+        expanded = _mcl_expand(mcl_adj, expansion)
         _mcl_inflate!(next_mcl_adj, expanded, inflation)
         _mcl_prune!(next_mcl_adj, prune_tol)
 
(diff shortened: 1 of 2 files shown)


src/mcl.jl
```diff
@@ -126,18 +105,18 @@ Identify clusters in the weighted graph using Markov Clustering Algorithm (MCL).
...
 """
-function mcl(adj::Matrix{Float64};
-             add_loops::Bool = true,
-             expansion::Number = 2, inflation::Number = 2.0,
-             save_final_matrix::Bool = false,
-             allow_singles::Bool = true,
-             max_iter::Integer = 100, tol::Number=1E-5,
-             display::Symbol=:none)#::MCLResult FIXME uncomment when 0.4 support is dropped
-
+function mcl{T<:Number}(adj::Matrix{T};
```
alyst: Can use this opportunity and change into `adj::AbstractMatrix{T}`

src/mcl.jl
```diff
@@ -126,18 +105,18 @@ Identify clusters in the weighted graph using Markov Clustering Algorithm (MCL).
   MCL doesn't converge
 * `max_iter::Integer`: max number of MCL iterations
 * `tol::Number`: MCL adjacency matrix convergence threshold
+# `prune_tol::Number`: Threshold pruning constant
```
alyst: maybe just "pruning threshold"?

src/mcl.jl
```diff
@@ -69,44 +69,23 @@ function _mcl_clusters(mcl_adj::Matrix{Float64}, allow_singles::Bool, zero_tol::
...
-# adjacency matrix inflation (element-wise raising to a given power) kernel
-function _mcl_inflate!(dest::Matrix{Float64}, src::Matrix{Float64}, inflation::Number)
-    any(src .< -1E-3*vecnorm(src)) && throw(InexactError())
-    map!(el -> max(0.0, el)^inflation, dest, src)
+# adjacency matrix pruning
+function _mcl_prune!(src::Matrix{Float64}, prune_tol::Number)
+    for i in 1:size(src,2)
+        c = view(src, :, i)
+        θ = mean(c)*prune_tol
+        c[c .< θ] = 0.0
```
alyst: IIUC, `c .< 0` will generate a temporary array (for every `i`), maybe something like
```julia
@inbounds @simd for j in eachindex(v)
    c[j] = ifelse(c[j] >= 0.0, c[j], 0.0)
end
```

src/mcl.jl
```diff
@@ -69,44 +69,23 @@ function _mcl_clusters(mcl_adj::Matrix{Float64}, allow_singles::Bool, zero_tol::
...
+function _mcl_inflate!{T<:Number}(dest::Matrix{Float64}, src::Matrix{T}, inflation::Number)
+    map!(el -> real((el+0im)^inflation), dest, src)
 end
 
-# adjacency matrix inflation (element-wise raising to a given power) kernel
-function _mcl_inflate!(dest::Matrix{Float64}, src::Matrix{Float64}, inflation::Number)
-    any(src .< -1E-3*vecnorm(src)) && throw(InexactError())
-    map!(el -> max(0.0, el)^inflation, dest, src)
+# adjacency matrix pruning
+function _mcl_prune!(src::Matrix{Float64}, prune_tol::Number)
```
alyst: Maybe rename `src` into `mtx` or `A` as it's modified here?

src/mcl.jl
```diff
@@ -69,44 +69,23 @@ function _mcl_clusters(mcl_adj::Matrix{Float64}, allow_singles::Bool, zero_tol::
...
-function _mcl_expand(src::Matrix, expansion::Number)
-    # we have to implement the workarond for matrix-power because of julia bug #16930
-    isinteger(expansion) && return _mcl_expand(src, Integer(real(expansion)))
-    v, X = eig(src)
-    # FIXME type instability here, revisit when #16930 is fixed
-    (eltype(v) <: Complex) || (any(v.<0) && (v = complex(v)))
-    Xinv = ishermitian(src) ? X' : inv(X)
-    X * Diagonal(v.^expansion) * Xinv
+# adjacency matrix inflation (element-wise raising to a given power) kernel
+function _mcl_inflate!{T<:Real}(dest::Matrix{Float64}, src::Matrix{T}, inflation::Integer)
```
alyst: `dest::Matrix{T}`?

src/mcl.jl
```diff
@@ -69,44 +69,23 @@ function _mcl_clusters(mcl_adj::Matrix{Float64}, allow_singles::Bool, zero_tol::
...
-    @inbounds for (i, el) in enumerate(src)
-        rel = real(el)
-        img = imag(el)
-        if rel < min_rel || (abs(img) > min_img && abs(img) > 1E-3*abs(rel))
-            throw(InexactError())
-        end
-        dest[i] = max(0.0, rel)^inflation
-    end
-    return dest
+function _mcl_inflate!{T<:Number}(dest::Matrix{Float64}, src::Matrix{T}, inflation::Number)
```
alyst: I would add here a comment that it's non-integral power inflation branch and what issue `real(el+0im)` fixes.
`dest::Matrix{T}`?

Actually, what about `el_inflate(x::Number, inflation::Integer)` and `el_inflate(x::Number, inflation::Number)` to avoid the duplication of the loop code?

src/mcl.jl
```diff
@@ -69,44 +69,23 @@ function _mcl_clusters(mcl_adj::Matrix{Float64}, allow_singles::Bool, zero_tol::
...
+    map!(el -> real((el+0im)^inflation), dest, src)
```
alyst: I think there are still some performance penalties when using anonymous functions:
```julia
x = rand(10000, 10000);
y = similar(x);

function f(y::Matrix, x::Matrix)
    @inbounds for i in eachindex(x)
        y[i] = max(x[i], 0)^2
    end
    return x
end

g(y::Matrix, x::Matrix) = (@inbounds map!(xx -> max(xx, 0)^2, y, x); y)
```

v0.6 timings (after few first runs) give
```
julia> @time f(y, x)
  0.158088 seconds (4 allocations: 160 bytes)
julia> @time f(y, x);
  0.162173 seconds (4 allocations: 160 bytes)
julia> @time f(y, x);
  0.155249 seconds (4 allocations: 160 bytes)
julia> @time f(y, x);
  0.155780 seconds (4 allocations: 160 bytes)
julia> @time f(y, x);
  0.160719 seconds (4 allocations: 160 bytes)
julia> @time f(y, x);
  0.158569 seconds (4 allocations: 160 bytes)
julia> @time f(y, x);
  0.158519 seconds (4 allocations: 160 bytes)
julia> @time f(y, x);
  0.164680 seconds (4 allocations: 160 bytes)

julia> @time g(y, x);
  0.240003 seconds (4 allocations: 160 bytes)
julia> @time g(y, x);
  0.226954 seconds (4 allocations: 160 bytes)
julia> @time g(y, x);
  0.233533 seconds (4 allocations: 160 bytes)
julia> @time g(y, x);
  0.232046 seconds (4 allocations: 160 bytes)
julia> @time g(y, x);
  0.225962 seconds (4 allocations: 160 bytes)
julia> @time g(y, x);
  0.230753 seconds (4 allocations: 160 bytes)
```

src/mcl.jl
alyst: However, as `inflation` is a keyarg parameter in `mcl()` and type inference of keyarg parameters would not be fixed until Julia v1.x, it's better not to call `el_inflate()` from `mcl()` directly.

test/mcl.jl
```diff
@@ -58,3 +58,16 @@ res = mcl(diagm([1.0, 1.0]), display=:none, allow_singles=false)
...
 @test res.assignments == [0, 0]
 @test res.nunassigned == 2
+
+# use sparse matrix
+res = mcl(sparse(adj_matrix), display=:none, expansion=2)
+@test isa(res, MCLResult)
+@test length(res.assignments) == length(nodes)
+@test res.nunassigned == 0
+
+@test_throws MethodError mcl(sparse(adj_matrix), display=:none, expansion=2.0)
```
alyst: Maybe `@test_broken` since it's nothing wrong in calling `mcl()` with these parameters, it's just that `^` is not implemented for sparse matrices yet?

alyst: Ah, please ignore the comment above -- I've just realized it doesn't make sense to implement `^non-integer` for sparse matrices.
It could make sense to check sparse `adj` + non-integer `expansion` combination in `mcl()` and throw `ArgumentError` with a more specific description, but `MethodError` should be fine for now.

src/mcl.jl
```diff
@@ -69,23 +69,35 @@ function _mcl_clusters(mcl_adj::Matrix{Float64}, allow_singles::Bool, zero_tol::
...
+function _mcl_prune!{T<:Number}(mtx::AbstractMatrix{T}, prune_tol::Number)
+    for i in 1:size(mtx,2)
+        c = view(mtx, :, i)
         θ = mean(c)*prune_tol
-        c[c .< θ] = 0.0
+        @inbounds @simd for j in eachindex(c)
+            c[j] = ifelse(c[j] >= θ, c[j], 0.0)
+        end
     end
+    issparse(mtx) && dropzeros!(mtx)
```
alyst: `return mtx`

src/mcl.jl
```diff
@@ -5,8 +5,8 @@
 
 Result returned by `mcl()`.
 """
-immutable MCLResult <: ClusteringResult
-    mcl_adj::Matrix{Float64}    # final MCL adjacency matrix (equilibrium state matrix if converged)
+immutable MCLResult{T<:Real} <: ClusteringResult
+    mcl_adj::AbstractMatrix{T}    # final MCL adjacency matrix (equilibrium state matrix if converged)
```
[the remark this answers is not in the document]
wildart: I do not think that it's important to make it concrete type.

src/mcl.jl
```diff
@@ -69,23 +69,35 @@ function _mcl_clusters(mcl_adj::Matrix{Float64}, allow_singles::Bool, zero_tol::
     clu_ixs, clu_sizes, unassigned_count
 end
 
-# adjacency matrix inflation (element-wise raising to a given power) kernel
-function _mcl_inflate!{T<:Real}(dest::Matrix{Float64}, src::Matrix{T}, inflation::Integer)
-    map!(el -> el^inflation, dest, src)
-end
+# adjacency matrix integral expansion (matrix-wise raising to a given power) kernel
+_mcl_expand{T<:Number}(src::AbstractSparseMatrix{T}, expansion::Integer) = src ^ expansion
```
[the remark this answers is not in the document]
wildart: There is no non-integral power function for sparse matrices. So method would and following declaration (below) would prevent usage of non-integral expansion factor and throw `MethodError` in `mcl` rather than somewhere in depths of `Base.LinAlg`.

src/mcl.jl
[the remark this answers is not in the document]
alyst: I agree, then it's better to drop `T`

test/mcl.jl
wildart: See my above comment, better to throw earlier.

src/mcl.jl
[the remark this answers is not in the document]
alyst: In fact, `speye(5)^2.0` works, so throwing `MethodError` for any non-Integer power might be too aggressive. Also this MethodError would not be overly descriptive. Maybe something like
```julia
function _mcl_expand(src::AbstractMatrix, expansion::Number)
    try
        return src^expansion
    catch ex
        if ex isa MethodError
            throw ArgumentError("MCL expansion of $(typeof(src)) with expansion=$expansion not supported")
        else
            rethrow()
        end
    end
end       
```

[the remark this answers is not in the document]
wildart: You're right.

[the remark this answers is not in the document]
alyst: I'm also a little bit concerned explicitly enumerating here what matrix types are supported, because the situation in Base may change over time.
For example, `Symmetric` matrices are supported just fine, but with the current two `_mcl_expand` methods they wouldn't be allowed.

wildart:
Ready to merge.

alyst:
LGTM, except I'm not sure we need to parameterize `MCLResult` on `T`, since it's user-facing yet doesn't provide any performance benefits. But if you think `T` could be useful, let's leave it. Also `T` in `_mcl_prune()` is not required.

wildart:
> not sure we need to parameterize MCLResult on T

True, no need for this. I removed unnecessary type parameters.

Issue #109, opened 2018-03-21
HClust initializes the Nearest Neighbor Array to 0-index leading to failure. 


sambitdash:
At places the code initializes the nearest neighbor array to 0-index leading to failure. This is particularly for the :single method when one of data points is not linked or a group of data points are not linked.

PR #106, opened 2017-12-11
hclust() refactoring + Ward method

alyst:
It started as a one line fix to reduce the number of allocations.
~~Actually, despite the big diff, the code of @davidavdav is quite efficient.~~ Now it's much faster, since it uses Lance-Williams approach for updating inter-cluster distances after clusters merging.

### Notable changes:
* Much faster `:average` and `:complete`
* `method` renamed to `linkage` (i.e. _cluster linkage_). It's less confusing (otherwise one may think it's about the actual algorithm to build the tree), also this is how it's called in e.g. Python's sklearn.
* Ward cluster linkage (`:ward` (=`Ward.D2` of R) and `:ward_presquared` (=`Ward.D1` of R)). It's implemented by `hclust_nn_lw()` internal method that accepts `metric::ReducibleMetric` param (_reducible_ Lance-Williams cluster-cluster distance). After the PR is landed it should be fairly straightforward to add the other Lance-Williams metrics (WPGMA, Centroid, Median -- the latter two are not reducible, but `hclust_nn_lw()` should still generate good approximation of the optimal tree)
* `Hclust` struct fields were changed (it carried too much R heritage): `height` => `heights`, `merge` => `merges`, `method` => `linkage`, `labels` field removed (there was no support for it anyway). The old field names are still available (via `getproperty()`) during the deprecation period.
* Internal refactoring to move the common code into separate structs/methods, e.g. internal `HclustMerges` and `HclustTrees` structs
* reduced GC overhead (in particular, by avoiding allocating 1-element vectors for the leaf nodes)
* the cluster numbers returned by `cutree()` match the R version
* `hclust_n3()` and `cutree()` covered by tests
* added rudimentary benchmark scripts
* fixes #109

## Benchmarks (the script is in `benchmark` folder):

| func   | param         | linkage  | time_old   | time_new   | gcratio_old | gcratio_new | allocs_old | allocs_new | memory_old | memory_new | n_old  | n_new |
|--------|---------------|----------|------------|------------|-------------|-------------|------------|------------|------------|------------|--------|-------|
| cutree | (10, 3)       | single   | 1.805 μs   | 11.412 μs  | 0.00%       | 0.00%       | 71         | 15         | 5.67 KiB   | 1.16 KiB   | 10000  | 902   |
| cutree | (100, 10)     | single   | 21.381 μs  | 21.040 μs  | 0.00%       | 0.00%       | 790        | 77         | 75.94 KiB  | 11.36 KiB  | 10000  | 938   |
[...]

PR #110, opened 2018-03-21
When disconnected node is presented in HClust(d, :single) 0-indexing does not kick-in

sambitdash:
Fix for #109


sambitdash:
The failure related to the 0.5 version is not related to the code checked-in.

alyst:
Thanks for the PR! The fixes for #109 should be in #106, which I just merged. I've also included your "isolated node" example from here.

PR #112, opened 2018-03-26
added Chinese Whisphers clustering

oxinabox:
This adds Chinese Whisphers clustering.
It works, the code could no doubt be better.

I've not done any docs yet.
Here is the paper it is from:
https://pdfs.semanticscholar.org/3e71/0251cb01ba6e1c0c735591776a212edc461f.pdf

This won't work until https://github.com/JuliaCollections/DataStructures.jl/pull/375
is merged and tagged.

It use the abstractions I suggested in 
https://github.com/JuliaLang/julia/issues/26613#issuecomment-376040464
This kinda spare matrix handling is I think a good for graph clustering algorithms

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,3 +3,4 @@ Compat 0.17
 Distances 0.3.1
 NearestNeighbors 0.0.3
 StatsBase 0.9.0
+DataStructures # need a newer as yet un released version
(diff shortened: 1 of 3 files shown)


ararslan:
Looks like you haven't included the file here.

oxinabox:
Oops, fixed.

PR #122, opened 2018-08-13
More Julia v0.7/1.0 fixes

alyst:
Also convert tests to use `@testset`


alyst:
I'm not sure why coverage drop is reported.
Will merge it tomorrow, unless there are objections.

PR #129, opened 2018-10-02
Support AbstractMatrix in fuzzy_cmeans

timholy:
The same change could be made in many other methods (looks like this package has lots of legacy code), but at least this is a step in the right direction.


timholy:
This also adds a convenient `show` method which can be used in doctests.

alyst:
Thanks! Could you please modify the `show()` method to report the number of iterations also when there's no convergence?

timholy:
Done

timholy:
Thanks!

I know it's a trivial change, but could you tag `v0.12.1`? ImageSegmentation needs this feature.

alyst:
v0.12.1 should be in the registry now.

Issue #131, opened 2018-11-09
affinity propagation result is not consistent with sklearn in python


jingpengw:
the two clustering results are different. Julia version did not do  any clustering since the assignment is just the index of each object! My similarity matrix is too large to show here.
```
using Clustering

@time affinityPropResult = Clustering.affinityprop(similarityMatrix)

affinityPropResult.assignments
```

```
using PyCall

@pyimport sklearn.cluster as cl
af = cl.AffinityPropagation(affinity="precomputed")[:fit](similarityMatrix)

labels = af[:labels_]
```

The travis test also did not verify the correctness of the result.

alyst:
Thanks for the report! You are most welcome to submit a fix.
Otherwise, if you have a small (<100 entities, the smaller the better) reproducible example, I can look into that.

jingpengw:
I double checked the results, they are same with some random tests.
I did not add  1 for python result since python start from 0.

labels = af[:labels_] .+ 1

jingpengw:
this issue still exist. The tests shows that it requires that the diagnal value should be the median value of the similarity matrix, otherwise the result is not consistent with python!

This is my test code:
```julia
using Distances
using Clustering
using LinearAlgebra
using Random
using Statistics 
Random.seed!(123)

    d = 10
    n = 44
    x = rand(d, n)
    S = -pairwise(Euclidean(), x, x)

    # set diagonal value to median value
#     S = S - diagm(0 => diag(S)) + median(S)*I

    R = affinityprop(S)


    k = length(R.exemplars)
    

using PyCall

@pyimport sklearn.cluster as cl
af = cl.AffinityPropagation(affinity="precomputed")[:fit]( S )
ref_assignments = af[:labels_] .+ 1

@assert randindex(R.assignments, ref_assignments)[2]==1.0
```

PR #132, opened 2018-11-19
affinity propagation: add comparison test with sklearn implementation

jingpengw:
see issue:
https://github.com/JuliaStats/Clustering.jl/issues/131


jingpengw:
this implementation seems more clear:
http://www.sc.ehu.es/ccwbayes/members/rsantana/software/matlab/Mateda2.0/otherfiles/apcluster.html

I'll try to reimplement it if I find some time.

alyst:
@jingpengw Thanks for checking this! I think it makes sense to merge with minor corrections:
 1. Include your PyCall script as a block comment (#= =#) before `python_assignments` with the comment that this is how the vector of assignments is generated: if later the test is modified, we would know how to update the expected output
 2. Rename `pythons_assignments` to `ref_assignments`, remove `labels` var; I guess it's not used/needed.
 3. Split the `ref_assignments` vector into several lines not exceeding 80 chars
 4. Remove all `println()` calls: they generate noise in CI, and if the `randindex()` test will fail, there would be actual-expected error log anyway

jingpengw:
updated, please take a look.

Issue #136, opened 2019-01-23
Make kmeans and kmeans! take AbstractMatrix instead of Matrix


tlienart:
At the moment feeding `kmeans` (and I suspect other algorithms, I haven't tried) with an adjoint will fail. You have to materialize the transpose which is wasteful.
Given that the convention `n x p` vs `p x n` depends heavily on the community (e.g. in ML it's very often `n x p` it seems to me), it would be good to allow this.

It seems to me that all is required is to change the definition of `kmeans` and `kmeans!` to `X::AbstractMatrix{T}`. I'm happy to open a PR for this if deemed worthy.

```julia
using Clustering
X = randn(50, 10)
kmeans(X', 2) # fails
kmeans(collect(X'), 2) # works but is ugly
```

Thanks!

alyst:
AFAIU `n x p` vs `p x n` is driven by the memory layout of a language. Julia is column-major, meaning that elements in the _i_-th column (`A[:, i]`) occupy continuous memory range. For many clustering algorithms it's more efficient when data items (rather than data dimensions) occupy continuous memory blocks. I think it's also the case for k-means as it uses `pairwise()` to calculate the distances between the elements.

So you can try allowing `AbstractMatrix`, but be sure to compare the performance for the matrices of different sizes (e.g. using `@time` or `@btime` from [BenchmarkTools.jl](https://github.com/JuliaCI/BenchmarkTools.jl)).

tlienart:
I understand the performance element of the convention choice, I'm not questioning that. Rather I'm just saying that it's highly likely other users would end up having a `X` in the "wrong" format (after all it's the convention in a large chunk of the literature so I don't think it's unreasonable to imagine the use case would arise) and having to then explicitly `collect(transpose(X))` which is both wasteful and a bit silly.

I think benchmarks are irrelevant for this, there may be an overhead for using AbstractMatrix but it would be completely dominated by the time needed to form the explicit transpose of `X` earlier and just benchmarking `kmeans` would hide that.

Anyway this exact problem popped up in MultivariateStats (I'm writing interfaces for a ML library) so I guess I'll just submit PRs and we'll see what happens.

alyst:
> Rather I'm just saying that it's highly likely other users would end up having a X in the "wrong" format ... and having to then explicitly `collect(transpose(X))` is both wasteful and a bit silly.

I'd rather recommend the users to adopt the optimal memory layout for serious applications.
[...]

PR #138, opened 2019-01-25
making kmeans take AbstractMatrix instead of Matrix

tlienart:
See #136 and #129. 

I duplicated the existing tests to take effectively the same matrix but now wrapped in a `Transpose` object.

Benchmarks on 1.1 / libopenlibm for what it's worth

**Before**

```julia
julia> seed!(1234)
julia> X = randn(5000, 200);
julia> @btime kmeans($X, 2);
3.138 ms (35 allocations: 95.34 KiB)
julia> @btime kmeans(transpose($Xt), 2);          # ERRORS
julia> @btime kmeans(copy(transpose($Xt)), 2);
4.835 ms (37 allocations: 7.72 MiB)
```

**After**
```julia
julia> seed!(1234)
julia> X = randn(5000, 200);
julia> @btime kmeans($X, 2);
3.158 ms (35 allocations: 95.34 KiB)
julia> @btime kmeans(transpose($Xt), 2);     # WORKS
21.913 ms (54 allocations: 96.10 KiB)
julia> @btime kmeans(copy(transpose($Xt)), 2);
  4.782 ms (37 allocations: 7.72 MiB)
```

So in short, the performances are identical, the only difference is that now it works if the user passes a transpose which I think is desirable. 
Of course we can argue that there is a performance drop if that's the case and it could be suggested that the user should do `copy(transpose(X))` if they can. My opinion is that were we should suggest it we should not force it and let the user decide whether that's a good idea for them or not. (the copy could be problematic if the data is very large).

### DIFF
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -243,7 +243,8 @@ function update_centers!(
     centers::AbstractMatrix{<:Real}, # out: updated centers (d x k)
     cweights::Vector{Float64})       # out: updated cluster weights (k)
 
-    (d, n), k = size(X), size(centers, 2)
+    d, n = size(X)
+    k = size(centers, 2)
 
     # initialize center weights
     cweights[to_update] .= 0.0
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -2,7 +2,7 @@
 
 #### Interface
 
-mutable struct KmeansResult{T<:AbstractFloat} <: ClusteringResult
+mutable struct KmeansResult{T<:Real} <: ClusteringResult
     centers::Matrix{T}         # cluster centers (d x k)
     assignments::Vector{Int}   # assignments (n)
     costs::Vector{Float64}     # costs of the resultant assignments (n)
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -34,7 +34,7 @@ function kmeans!(X::AbstractMatrix{<:Real}, centers::AbstractMatrix{<:Real};
     counts = Vector{Int}(undef, k)
     cweights = Vector{Float64}(undef, k)
 
-    _kmeans!(X, conv_weights(Float64, n, weights), centers, assignments,
+    _kmeans!(X, conv_weights(eltype(X), n, weights), centers, assignments,
              costs, counts, cweights, maxiter, tol,
              display_level(display), distance)
 end
@@ -62,9 +62,9 @@ end
 
 # core k-means skeleton
 function _kmeans!(
-    X::AbstractMatrix{<:Real},        # in: sample matrix (d x n)
+    X::AbstractMatrix{<:Real},                 # in: sample matrix (d x n)
     w::Union{Nothing, AbstractVector{<:Real}}, # in: sample weights (n)
-    centers::AbstractMatrix{<:Real},  # in/out: matrix of centers (d x k)
+    centers::AbstractMatrix{<:Real},           # in/out: matrix of centers (d x k)
     assignments::Vector{Int},     # out: vector of assignments (n)
     costs::Vector{Float64},       # out: costs of the resultant assignments (n)
     counts::Vector{Int},          # out: # samples assigned to each cluster (k)
@@ -82,10 +82,8 @@ function _kmeans!(
     num_affected::Int = k # number of centers, to which the distances need to be recomputed
 
     dmat = pairwise(distance, centers, X)
-    update_assignments!(dmat, true, assignments, costs, counts,
-                        to_update, unused)
-
-    objv = (w === nothing) ? sum(costs) : dot(w, costs)
+    update_assignments!(dmat, true, assignments, costs, counts, to_update, unused)
+    objv = w === nothing ? sum(costs) : dot(w, costs)
 
     # main loop
     t = 0
@@ -122,14 +120,13 @@ function _kmeans!(
 
         # update assignments
 
-        update_assignments!(dmat, false, assignments, costs, counts,
-                            to_update, unused)
+        update_assignments!(dmat, false, assignments, costs, counts, to_update, unused)
 
         num_affected = sum(to_update) + length(unused)
 
         # compute change of objective and determine convergence
         prev_objv = objv
-        objv = (w === nothing) ? sum(costs) : dot(w, costs)
+        objv = w === nothing ? sum(costs) : dot(w, costs)
         objv_change = objv - prev_objv
 
         if objv_change > tol
@@ -165,13 +162,13 @@ end
 #  an updated (squared) distance matrix
 #
 function update_assignments!(
-    dmat::Matrix{<:Real},     # in:  distance matrix (k x n)
-    is_init::Bool,            # in:  whether it is the initial run
-    assignments::Vector{Int}, # out: assignment vector (n)
-    costs::Vector{Float64},   # out: costs of the resultant assignment (n)
-    counts::Vector{Int},      # out: # samples assigned to each cluster (k)
-    to_update::Vector{Bool},  # out: whether a center needs update (k)
-    unused::Vector{Int})      # out: list of centers with no samples
+    dmat::AbstractMatrix{<:Real}, # in:  distance matrix (k x n)
+    is_init::Bool,                # in:  whether it is the initial run
+    assignments::Vector{Int},     # out: assignment vector (n)
+    costs::Vector{Float64},       # out: costs of the resultant assignment (n)
+    counts::Vector{Int},          # out: # samples assigned to each cluster (k)
+    to_update::Vector{Bool},      # out: whether a center needs update (k)
+    unused::Vector{Int})          # out: list of centers with no samples
 
     k, n = size(dmat)
 
@@ -255,14 +252,11 @@ function update_centers!(
         1 <= cj <= k || error("assignment out of boundary.")
 
         if to_update[cj]
+            xj = view(X, :, j)
             if cweights[cj] > 0
-                centers[:, cj] .+= view(X, :, j)
-                    centers[i, cj] += X[i, j]
-                end
+                centers[:, cj] .+= xj
             else
-                centers[:, cj] .= view(X, :, j)
-                    centers[i, cj] = X[i, j]
-                end
+                centers[:, cj] .= xj
             end
             cweights[cj] += 1
         end
@@ -271,11 +265,7 @@ function update_centers!(
     # sum ==> mean
     @inbounds for j = 1:k
         if to_update[j]
-            centers[:, j] .*= 1.0 / cweights[j]
-            vj = view(centers, :, j)
-            for i = 1:d
-                vj[i] *= cj
-            end
+            centers[:, j] .*= 1 / cweights[j]
         end
     end
 end
@@ -300,20 +290,18 @@ function update_centers!(
     cweights[to_update] .= 0.0
 
     # accumulate columns
-    # accumulate_cols_u!(centers, cweights, x, assignments, weights, to_update)
-    for j = 1:n
-        @inbounds wj = weights[j]
+    @inbounds for j = 1:n
+        wj = weights[j]
         if wj > 0
-            @inbounds cj = assignments[j]
+            cj = assignments[j]
             1 <= cj <= k || error("assignment out of boundary.")
 
             if to_update[cj]
-                rj = view(centers, :, cj)
                 xj = view(X, :, j)
                 if cweights[cj] > 0
-                    @inbounds rj .+= xj * wj
+                    centers[:, cj] .+= xj * wj
                 else
-                    rj .= xj * wj
+                    centers[:, cj] .= xj * wj
                 end
                 cweights[cj] += wj
             end
@@ -321,9 +309,9 @@ function update_centers!(
     end
 
     # sum ==> mean
-    for j = 1:k
+    @inbounds for j = 1:k
         if to_update[j]
-            @inbounds centers[:, j] .*= 1.0 / cweights[j]
+            centers[:, j] .*= 1 / cweights[j]
         end
     end
 end
@@ -336,7 +324,7 @@ function repick_unused_centers(
     X::AbstractMatrix{<:Real},       # in: the sample set (d x n)
     costs::Vector{Float64},          # in: the current assignment costs (n)
     centers::AbstractMatrix{<:Real}, # to be updated: the centers (d x k)
-    unused::Vector{Int})          # in: set of indices of centers to be updated
+    unused::Vector{Int})             # in: set of indices of centers to be updated
 
     # pick new centers using a scheme like kmeans++
     ds = similar(costs)
@@ -346,7 +334,7 @@ function repick_unused_centers(
     for i in unused
         j = wsample(1:n, tcosts)
         tcosts[j] = 0
-        v = X[:, j]
+        v = view(X, :, j)
         centers[:, i] = v
 
         colwise!(ds, SqEuclidean(), v, X)
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -5,7 +5,7 @@
 mutable struct KmeansResult{T<:AbstractFloat} <: ClusteringResult
     centers::Matrix{T}         # cluster centers (d x k)
     assignments::Vector{Int}   # assignments (n)
-    costs::Vector{T}           # costs of the resultant assignments (n)
+    costs::Vector{Float64}     # costs of the resultant assignments (n)
     counts::Vector{Int}        # number of samples assigned to each cluster (k)
     cweights::Vector{Float64}  # cluster weights (k)
     totalcost::Float64         # total cost (i.e. objective) (k)
@@ -18,34 +18,33 @@ const _kmeans_default_maxiter = 100
 const _kmeans_default_tol = 1.0e-6
 const _kmeans_default_display = :none
 
-function kmeans!(X::AbstractMatrix{T}, centers::AbstractMatrix{T};
+function kmeans!(X::AbstractMatrix{<:Real}, centers::AbstractMatrix{<:Real};
                  weights=nothing, maxiter::Integer=_kmeans_default_maxiter,
                  tol::Real=_kmeans_default_tol,
                  display::Symbol=_kmeans_default_display,
-                 distance::SemiMetric=SqEuclidean()) where T<:AbstractFloat
+                 distance::SemiMetric=SqEuclidean())
 
     m, n = size(X)
     m2, k = size(centers)
     m == m2 || throw(DimensionMismatch("Inconsistent array dimensions."))
     (2 <= k < n) || error("k must have 2 <= k < n.")
 
-    assignments = zeros(Int, n)
-    costs = zeros(T, n)
+    assignments = Vector{Int}(undef, n)
+    costs = Vector{Float64}(undef, n)
     counts = Vector{Int}(undef, k)
     cweights = Vector{Float64}(undef, k)
 
-    _kmeans!(X, conv_weights(T, n, weights), centers,
-             assignments, costs, counts, cweights,
-             round(Int, maxiter), tol, display_level(display), distance)
+    _kmeans!(X, conv_weights(Float64, n, weights), centers, assignments,
+             costs, counts, cweights, maxiter, tol,
+             display_level(display), distance)
 end
 
-function kmeans(X::AbstractMatrix{T}, k::Int;
-                weights=nothing,
-                init=_kmeans_default_init,
+function kmeans(X::AbstractMatrix{<:Real}, k::Int;
+                weights=nothing, init=_kmeans_default_init,
                 maxiter::Integer=_kmeans_default_maxiter,
                 tol::Real=_kmeans_default_tol,
                 display::Symbol=_kmeans_default_display,
-                distance::SemiMetric=SqEuclidean()) where T<:AbstractFloat
+                distance::SemiMetric=SqEuclidean())
 
     m, n = size(X)
     (2 <= k < n) || error("k must have 2 <= k < n.")
@@ -63,30 +62,30 @@ end
 
 # core k-means skeleton
 function _kmeans!(
-    x::AbstractMatrix{T},         # in: sample matrix (d x n)
-    w::Union{Nothing, Vector{T}}, # in: sample weights (n)
-    centers::AbstractMatrix{T},   # in/out: matrix of centers (d x k)
+    X::AbstractMatrix{<:Real},        # in: sample matrix (d x n)
+    w::Union{Nothing, AbstractVector{<:Real}}, # in: sample weights (n)
+    centers::AbstractMatrix{<:Real},  # in/out: matrix of centers (d x k)
     assignments::Vector{Int},     # out: vector of assignments (n)
-    costs::Vector{T},             # out: costs of the resultant assignments (n)
-    counts::Vector{Int},          # out: the number of samples assigned to each cluster (k)
-    cweights::Vector{Float64},    # out: the weights of each cluster
-    maxiter::Int,                 # in: maximum number of iterations
+    costs::Vector{Float64},       # out: costs of the resultant assignments (n)
+    counts::Vector{Int},          # out: # samples assigned to each cluster (k)
+    cweights::Vector{Float64},    # out: weights of each cluster
+    maxiter::Integer,             # in: maximum number of iterations
     tol::Real,                    # in: tolerance of change at convergence
     displevel::Int,               # in: the level of display
     distance::SemiMetric          # in: function to calculate the distance with
-    ) where T<:AbstractFloat
+    )
 
     # initialize
-
     k = size(centers, 2)
     to_update = Vector{Bool}(undef, k) # indicators of whether a center needs to be updated
-    unused = Int[]
+    unused = Vector{Int}()
     num_affected::Int = k # number of centers, to which the distances need to be recomputed
 
-    dmat = pairwise(distance, centers, x)
-    dmat = convert(Array{T}, dmat) #Can be removed if one day Distance.result_type(SqEuclidean(), T, T) == T
-    update_assignments!(dmat, true, assignments, costs, counts, to_update, unused)
-    objv = w === nothing ? sum(costs) : dot(w, costs)
+    dmat = pairwise(distance, centers, X)
+    update_assignments!(dmat, true, assignments, costs, counts,
+                        to_update, unused)
+
+    objv = (w === nothing) ? sum(costs) : dot(w, costs)
 
     # main loop
     t = 0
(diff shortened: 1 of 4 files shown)


src/kmeans.jl
```diff
@@ -64,17 +63,18 @@ end
...
-    assignments::Vector{Int},       # out: vector of assignments (n)
-    costs::Vector{T},               # out: costs of the resultant assignments (n)
-    counts::Vector{Int},            # out: the number of samples assigned to each cluster (k)
-    cweights::Vector{Float64},      # out: the weights of each cluster
-    maxiter::Int,                   # in: maximum number of iterations
-    tol::Real,                      # in: tolerance of change at convergence
-    displevel::Int,                 # in: the level of display
-    distance::SemiMetric) where T<:AbstractFloat             # in: function to calculate the distance with
+    x::AbstractMatrix{T},         # in: sample matrix (d x n)
+    w::Union{Nothing, Vector{T}}, # in: sample weights (n)
```
alyst: It would also be nice to change `Vector{T}` (and `Vector{Int}` etc) into `AbstractVector{T}` (resp. `AbstractVector{Int}`)  here and in the other functions.

src/kmeans.jl
```diff
@@ -64,17 +63,18 @@ end
...
+    centers::AbstractMatrix{T},   # in/out: matrix of centers (d x k)
+    assignments::Vector{Int},     # out: vector of assignments (n)
+    costs::Vector{T},             # out: costs of the resultant assignments (n)
+    counts::Vector{Int},          # out: the number of samples assigned to each cluster (k)
+    cweights::Vector{Float64},    # out: the weights of each cluster
```
alyst: `AbstractVector{<:AbstractFloat}`

src/kmeans.jl
```diff
@@ -64,17 +63,18 @@ end
...
+    maxiter::Int,                 # in: maximum number of iterations
```
alyst: `Integer`

alyst:
Thanks! Besides `Transpose` there are other `AbstractMatrix` subtypes, which may work well without `collect()` (in particular, `SubArray`), so these changes would be quite useful.

src/kmeans.jl
tlienart: Ok thanks @alyst , I'll try to apply the changes in a way that's easily reviewable and if it looks like it might be too much then maybe another PR would be better.

tlienart: Actually I'm not sure that makes a lot of sense given that the caller function allocates those vectors explicitly?

tlienart:
I made all the changes requested by @nalimilan and some of the changes requested by @alyst. 

Re @alyst: since in `kmeans!` there are now effectively those lines:

```julia
assignments = Vector{Int}(undef, n)
costs = Vector{Float64}(undef, n)
counts = Vector{Int}(undef, k)
cweights = Vector{Float64}(undef, k)
```

I don't think it's useful to have subsequent helper functions take `AbstractVector` etc. I tried to make changes that I thought would help towards generality though and, where appropriate, tried to be explicit about the type if the code warrants it. I hope the changes are ok. 

(on cosmetics, I hope you won't mind too much, I was just reviewing my changes and ended up doing a bit of cleaning up...)

src/kmeans.jl
```diff
@@ -5,7 +5,7 @@
 mutable struct KmeansResult{T<:AbstractFloat} <: ClusteringResult
```
alyst: Since `kmeans()` functions now accept `AbstractMatrix{<:Real}`, I guess here it also should be relaxed to `T<:Real`

src/kmeans.jl
```diff
@@ -238,49 +236,44 @@ end
...
+    w::Nothing,                      # in: sample weights
+    assignments::Vector{Int},        # in: assignments (n)
+    to_update::Vector{Bool},         # in: whether a center needs update (k)
+    centers::AbstractMatrix{<:Real}, # out: updated centers (d x k)
+    cweights::Vector{Float64})       # out: updated cluster weights (k)
 
-    d::Int = size(x, 1)
-    n::Int = size(x, 2)
-    k::Int = size(centers, 2)
+    (d, n), k = size(X), size(centers, 2)
```
alyst: I also like terse syntax, but this one looks a bit exotic too me. :) I'd rather use
```julia
d, n = size(X)
k = size(centers, 2)
```

src/kmeans.jl
alyst: Since `kmeans()` functions now accept `AbstractMatrix{<:Real}`, I guess here it also should be relaxed

```suggestion
mutable struct KmeansResult{T<:Real} <: ClusteringResult
```

src/kmeans.jl
alyst: I also like terse syntax, but this one looks a bit exotic too me. :)

```suggestion
    d, n = size(X)
    k = size(centers, 2)
```

src/kmeans.jl
```diff
@@ -238,49 +236,44 @@ end
...
 
     # accumulate columns
-    @inbounds for j = 1 : n
+    @inbounds for j = 1:n
         cj = assignments[j]
         1 <= cj <= k || error("assignment out of boundary.")
+
         if to_update[cj]
             if cweights[cj] > 0
                 for i = 1:d
```
alyst: While you're on it, and since you're already doing these kind of changes (don't know whether it's possible to do multi-line suggestions)
```suggestion
                centers[:, cj] .+= view(X, :, j)
```

src/kmeans.jl
```diff
@@ -238,49 +236,44 @@ end
...
-                    centers[i, cj] += x[i, j]
+                    centers[i, cj] += X[i, j]
                 end
             else
                 for i = 1:d
```
```suggestion
                centers[:, cj] .= view(X, :, j)
```

src/kmeans.jl
```diff
@@ -238,49 +236,44 @@ end
...
         end
     end
 
     # sum ==> mean
-    for j = 1:k
+    @inbounds for j = 1:k
         if to_update[j]
-            @inbounds cj::T = 1 / cweights[j]
-            vj = view(centers,:,j)
+            cj = 1.0 / cweights[j]
```
```suggestion
            centers[:, j] .*= 1.0 / cweights[j]
```

src/kmeans.jl
```diff
@@ -292,37 +285,33 @@ end
...
-    n::Int = size(x, 2)
-    k::Int = size(centers, 2)
+    X::AbstractMatrix{<:Real},       # in: sample matrix (d x n)
+    weights::AbstractVector{<:Real}, # in: sample weights (n)
+    assignments::Vector{Int},        # in: assignments (n)
+    to_update::Vector{Bool},         # in: whether a center needs update (k)
+    centers::AbstractMatrix{<:Real}, # out: updated centers (d x k)
+    cweights::Vector{Float64})       # out: updated cluster weights (k)
+
+    (d, n), k = size(X), size(centers, 2)
```
```suggestion
    d, n = size(X)
    k = size(centers, 2)
```

src/kmeans.jl
```diff
@@ -292,37 +285,33 @@ end
...
 
             if to_update[cj]
                 rj = view(centers, :, cj)
-                xj = view(x, :, j)
+                xj = view(X, :, j)
                 if cweights[cj] > 0
                     @inbounds rj .+= xj * wj
                 else
-                    @inbounds rj .= xj * wj
+                    rj .= xj * wj
```
alyst: Are you sure `@inbounds` is not required here?

src/kmeans.jl
```diff
@@ -342,23 +331,23 @@ end
...
     tcosts = copy(costs)
-    n = size(x, 2)
+    n = size(X, 2)
 
     for i in unused
         j = wsample(1:n, tcosts)
         tcosts[j] = 0
-        v = x[:,j]
-        centers[:,i] = v
+        v = X[:, j]
```
```suggestion
        v = view(X, :, j)
```

src/seeding.jl
```diff
@@ -53,8 +54,8 @@ function copyseeds!(S::DenseMatrix, X::DenseMatrix, iseeds::AbstractVector)
     return S
 end
 
-copyseeds(X::DenseMatrix{T}, iseeds::AbstractVector) where {T} =
-    copyseeds!(Matrix{T}(undef, size(X,1), length(iseeds)), X, iseeds)
+copyseeds(X::AbstractMatrix{<:Real}, iseeds::AbstractVector) =
+    copyseeds!(Matrix{eltype(X)}(undef, size(X,1), length(iseeds)), X, iseeds)
```
alyst:
```suggestion
    copyseeds!(similar(X, size(X, 1), length(iseeds)), X, iseeds)
```

src/seeding.jl
```diff
@@ -128,7 +131,7 @@ function initseeds_by_costs!(iseeds::IntegerVector, alg::KmppAlg, costs::RealMat
     iseeds[1] = p
 
     if k > 1
-        mincosts = copy(view(costs,:,p))
+        mincosts = copy(view(costs, :, p))
```
alyst:
```suggestion
        mincosts = costs[:, p]
```

src/kmeans.jl
```diff
@@ -18,35 +18,33 @@ const _kmeans_default_maxiter = 100
...
-    costs = zeros(T, n)
+    assignments = Vector{Int}(undef, n)
+    costs = Vector{Float64}(undef, n)
     counts = Vector{Int}(undef, k)
     cweights = Vector{Float64}(undef, k)
 
-    _kmeans!(X, conv_weights(T, n, weights), centers,
-             assignments, costs, counts, cweights,
-             round(Int, maxiter), tol, display_level(display), distance)
+    _kmeans!(X, conv_weights(Float64, n, weights), centers, assignments,
```
nalimilan: `T` was actually needed in this function, otherwise it can become type unstable.

src/kmeans.jl
```diff
@@ -64,29 +62,30 @@ end
...
-    centers::Matrix{T},             # in/out: matrix of centers (d x k)
-    assignments::Vector{Int},       # out: vector of assignments (n)
-    costs::Vector{T},               # out: costs of the resultant assignments (n)
-    counts::Vector{Int},            # out: the number of samples assigned to each cluster (k)
-    cweights::Vector{Float64},      # out: the weights of each cluster
-    maxiter::Int,                   # in: maximum number of iterations
-    tol::Real,                      # in: tolerance of change at convergence
-    displevel::Int,                 # in: the level of display
-    distance::SemiMetric) where T<:AbstractFloat             # in: function to calculate the distance with
+    X::AbstractMatrix{<:Real},        # in: sample matrix (d x n)
```
nalimilan: Broken alignment.

src/kmeans.jl
```diff
@@ -64,29 +62,30 @@ end
...
-    unused = Int[]
+    unused = Vector{Int}()
     num_affected::Int = k # number of centers, to which the distances need to be recomputed
 
-    dmat = pairwise(distance, centers, x)
-    dmat = convert(Array{T}, dmat) #Can be removed if one day Distance.result_type(SqEuclidean(), T, T) == T
-    update_assignments!(dmat, true, assignments, costs, counts, to_update, unused)
-    objv = w === nothing ? sum(costs) : dot(w, costs)
+    dmat = pairwise(distance, centers, X)
+    update_assignments!(dmat, true, assignments, costs, counts,
```
nalimilan: Why change these two lines? They were fine previously. Same below.

src/kmeans.jl
```diff
@@ -64,29 +62,30 @@ end
...
     # initialize
-
     k = size(centers, 2)
     to_update = Vector{Bool}(undef, k) # indicators of whether a center needs to be updated
-    unused = Int[]
+    unused = Vector{Int}()
     num_affected::Int = k # number of centers, to which the distances need to be recomputed
 
-    dmat = pairwise(distance, centers, x)
-    dmat = convert(Array{T}, dmat) #Can be removed if one day Distance.result_type(SqEuclidean(), T, T) == T
```
nalimilan: Are you sure this is the case now?

src/kmeans.jl
nalimilan: Should keep `1` instead of `1.0`, as it preserves the type.

src/kmeans.jl
```diff
@@ -332,7 +321,7 @@ function update_centers!(
     # sum ==> mean
     for j = 1:k
         if to_update[j]
-            @inbounds centers[:, j] .*= 1 / cweights[j]
+            @inbounds centers[:, j] .*= 1.0 / cweights[j]
```
nalimilan: Same.

src/kmeans.jl
```diff
@@ -342,23 +331,23 @@ end
...
 #
 function repick_unused_centers(
-    x::Matrix{T},           # in: the sample set (d x n)
-    costs::Vector{T},       # in: the current assignment costs (n)
-    centers::Matrix{T},     # to be updated: the centers (d x k)
-    unused::Vector{Int}) where T<:AbstractFloat    # in: the set of indices of centers to be updated
+    X::AbstractMatrix{<:Real},       # in: the sample set (d x n)
+    costs::Vector{Float64},          # in: the current assignment costs (n)
+    centers::AbstractMatrix{<:Real}, # to be updated: the centers (d x k)
+    unused::Vector{Int})          # in: set of indices of centers to be updated
```
nalimilan: Broken alignment.

src/seeding.jl
```diff
@@ -31,16 +31,17 @@ seeding_algorithm(s::Symbol) =
...
+initseeds_by_costs(algname::Symbol, costs::AbstractMatrix{<:Real}, k::Integer) =
     initseeds_by_costs(seeding_algorithm(algname), costs, k)
 
-initseeds(iseeds::Vector{Int}, X::RealMatrix, k::Integer) = iseeds
-initseeds_by_costs(iseeds::Vector{Int}, costs::RealMatrix, k::Integer) = iseeds
+initseeds(iseeds::Vector{Int}, X::AbstractMatrix{<:Real}, k::Integer) = iseeds
+initseeds_by_costs(iseeds::Vector{Int}, costs::AbstractMatrix{<:Real}, k::Integer) = iseeds
 
-function copyseeds!(S::DenseMatrix, X::DenseMatrix, iseeds::AbstractVector)
+function copyseeds!(S::DenseMatrix, X::AbstractMatrix{<:Real},
```
nalimilan: Old version was right.

src/seeding.jl
nalimilan: Then also stop using `DenseMatrix` in `copyseeds!`.

src/seeding.jl
```diff
@@ -69,11 +70,9 @@ end
 
 mutable struct RandSeedAlg <: SeedingAlgorithm end
 
-initseeds!(iseeds::IntegerVector, alg::RandSeedAlg, X::RealMatrix) =
```
nalimilan: Keep line break. Same below.

src/seeding.jl
```diff
@@ -85,7 +84,9 @@ initseeds_by_costs!(iseeds::IntegerVector, alg::RandSeedAlg, X::RealMatrix) =
 
 mutable struct KmppAlg <: SeedingAlgorithm end
 
-function initseeds!(iseeds::IntegerVector, alg::KmppAlg, X::RealMatrix, metric::PreMetric)
+function initseeds!(iseeds::IntegerVector, alg::KmppAlg,
+                    X::AbstractMatrix{<:Real}, metric::PreMetric)
+
```
nalimilan: Julia conventions generally don't use empty lines at top of functions.

src/seeding.jl
```diff
@@ -115,10 +116,12 @@ function initseeds!(iseeds::IntegerVector, alg::KmppAlg, X::RealMatrix, metric::
     return iseeds
 end
 
-initseeds!(iseeds::IntegerVector, alg::KmppAlg, X::RealMatrix) =
+initseeds!(iseeds::IntegerVector, alg::KmppAlg, X::AbstractMatrix{<:Real}) =
     initseeds!(iseeds, alg, X, SqEuclidean())
 
-function initseeds_by_costs!(iseeds::IntegerVector, alg::KmppAlg, costs::RealMatrix)
+function initseeds_by_costs!(iseeds::IntegerVector, alg::KmppAlg,
+                             costs::AbstractMatrix{<:Real})
```
nalimilan: Same.

test/seeding.jl
```diff
@@ -31,52 +31,91 @@ k = 5
...
+Ct = copy(transpose(C))
+
 md0 = min_interdist(X)
 
 @testset "RandSeed" begin
+    Random.seed!(34568)
     iseeds = initseeds(RandSeedAlg(), X, k)
     @test length(iseeds) == k
     @test alldistinct(iseeds)
+    Random.seed!(34568)
```
nalimilan: Is it really useful to repeat this? All that matters is that the result is the same across runs, and for this one call is enough.

src/kmeans.jl
tlienart: I put it back

src/kmeans.jl
tlienart: With apologies, I don't really understand this comment and more importantly I'm not sure what to do.
Reverting to the previous doesn't seem right to me as it imposes the type of `X` and `centers` to be the same which I don't think is correct. 
I've changed the `Float64` in the `conv_weights` to `eltype(X)` is that ok?

src/kmeans.jl
tlienart: `pairwise` takes an abstractmatrix, modifies it in place and returns it (https://github.com/JuliaStats/Distances.jl/blob/c4431c025790d11a85171e383e56b48f8d3d78c5/src/generic.jl#L84_L95 so I don't think there needs to be a conversion here. However `update_assignments` needs to take an `AbstractMatrix`, I've changed that.

test/seeding.jl
tlienart: Yes I believe so given that the seeding functions are all random, the tape needs to be rewinded if I want to test that the same numbers are generated.

tlienart:
Thanks for the detailed feedback. I think I applied all the changes requested apart from 2 where I'm not sure I understood the comment and may have not interpreted it correctly.

Apologies for the bloated PR I wasn't expecting to have to touch so many parts of the code 🙈

src/kmeans.jl
nalimilan: No, that's `pairwise!`. `pairwise` allocates a ` Matrix{result_type(metric, a, b)}`. Unless you really know what you are doing, better keep this, anyway it will be a no-op if the type is already correct.

tlienart: ok!

src/kmeans.jl
```diff
@@ -18,35 +18,33 @@ const _kmeans_default_maxiter = 100
...
 
     m, n = size(X)
     m2, k = size(centers)
     m == m2 || throw(DimensionMismatch("Inconsistent array dimensions."))
     (2 <= k < n) || error("k must have 2 <= k < n.")
 
-    assignments = zeros(Int, n)
-    costs = zeros(T, n)
+    assignments = Vector{Int}(undef, n)
+    costs = Vector{Float64}(undef, n)
```
nalimilan: Should use `eltype(X)` here too: using the same eltype as `X` can improve performance when using e.g. `Float32` and can help preserving that type in multiplications. Same for the definition of `KMeansResult` above.

src/kmeans.jl
nalimilan: `eltype(X)` is fine. You could also have kept `T`, just not using it for `centers`.

However, please check that the code (this function and others down the line) is type stable when `X` and `centers` have different eltypes: it may have been written with the assumption that they are equal.

src/kmeans.jl
tlienart: apologies for being a PITA here but what's the proper type? surely not elements of `X` or `centers`, it's unrelated and `dmat` is only assigned via `pairwise` and only used in the function `update_assignments!` where the only thing that matters is that its elements can be compared between themselves so anything `<:Real` is definitely fine.
So given we know `pairwise` returns a `Matrix` of a type `<:Real` and as long as `update_assignments!` takes exactly that, I'm not sure I see a point where this convert will be useful?

src/kmeans.jl
```diff
@@ -167,15 +162,15 @@ end
...
-    is_init::Bool,              # in:  whether it is the initial run
-    assignments::Vector{Int},   # out: assignment vector (n)
-    costs::Vector{T},           # out: costs of the resultant assignment (n)
-    counts::Vector{Int},        # out: number of samples assigned to each cluster (k)
-    to_update::Vector{Bool},    # out: whether a center needs update (k)
-    unused::Vector{Int}) where T<:AbstractFloat        # out: the list of centers get no samples assigned to it
+    dmat::AbstractMatrix{<:Real}, # in:  distance matrix (k x n)
+    is_init::Bool,                # in:  whether it is the initial run
+    assignments::Vector{Int},     # out: assignment vector (n)
+    costs::Vector{Float64},       # out: costs of the resultant assignment (n)
```
```suggestion
    costs::Vector{<:Real},       # out: costs of the resultant assignment (n)
```

src/kmeans.jl
tlienart: with hindsight, I'm not happy with this conversion. If the `X` are say, integers and the weights are float, the weights should stay floats. So it should be something like `conv_weights(promote_type(eltype(X), Float64), n, weights)`

src/kmeans.jl
nalimilan: TBH I don't really know, one would need to check the history and see whether the function is type stable under various combinations of types for `X` and `centers`. As I noted in another comment, if `eltype(X) === Float32`, computations can be faster if other matrices are also in `Float32` rather than `Float64`, so that may be a reason. Another one is that `dmat` is reused below and to avoid type unstability it must not change its type.

Anyway, a general rule is that when you don't understand why a line is there, either keep it as-is, or investigate until you understand why it's been added.

src/kmeans.jl
nalimilan: Good catch. But `promote_type(eltype(X), Float64)` will almost always return `Float64`. Better use `float(eltype(X))`. Or maybe better, use `eltype(centers)`, which cannot be integers, so that the user can choose the most appropriate type.

tlienart:
I've applied the changes in 16ae803. 

I must admit there are some things that I still don't like with the types (e.g. centers are updated with X*w, but X and c have the same type while w could potentially be something else) but this PR is already too messy so that would be for another issue.

src/kmeans.jl
alyst: > `T` was actually needed in this function, otherwise it can become type unstable.

Are the relevant tests already there (`@inferred` `typeof() ==` etc) or could they be added? Otherwise finding the right types is a little bit like wandering in the dark.

tlienart: I don't think so but maybe that could be for another PR also related to #140 so it's not just for `kmeans` that such checks would be needed...

src/kmeans.jl
```diff
@@ -295,13 +299,15 @@ function update_centers!(
         if wj > 0
             cj = assignments[j]
             1 <= cj <= k || error("assignment out of boundary.")
-
             if to_update[cj]
-                xj = view(X, :, j)
                 if cweights[cj] > 0
-                    centers[:, cj] .+= xj * wj
```
alyst: I guess this `*` (plus the one below) was the reason for regression in weighted case as it creates the new vector.
`.+= xj .* wj` should avoid that.
I'm a little bit hesitant about converting everything to explicit loops as it defeats the purpose of having broadcast support in the language.

src/kmeans.jl
```diff
@@ -311,7 +317,9 @@ function update_centers!(
     # sum ==> mean
     @inbounds for j = 1:k
         if to_update[j]
-            centers[:, j] .*= 1 / cweights[j]
+            for i = 1:d
+                centers[i, j] /= cweights[j]
```
alyst: I'm not 100% sure, but back in the days `/` was more expensive than `*`.

src/kmeans.jl
tlienart: I'm happy with opening another PR for this kind of stuff after this one gets merged so that we can isolate changes related to the initial problem from other changes?

tlienart: (FWIW, I tried exactly what you said and it increases the allocation by a lot and doubles the time, possibly because it constructs `centers[:, cj]`. 
There may be better things to do with `copyto!` etc but I must admit I'd rather get this PR done with first...

alyst: I thought `centers[:, cj] .+=` calls `setindex!()` directly, but it actually creates `centers[:, cj]` vector behind the scenes (`centers[:, cj] .=` doesn't). So the better version is `cej = view(centers, :, cj); cej .+= xj .* wj`.

src/kmeans.jl
nalimilan: Note that the current code computes `1/cweights[j]` only once, and then does the multiplication for each `i`. IIUC the timings for `*= 1 /` recompute the division it for each `i`, which isn't faster.

alyst: Actually, I think LLVM is smart enough to calculate `1/wj` only once (at least that's what I see in `@code_llvm` for a toy example). But I still prefer the broadcast version (`cej = view(centers, :, j); cej .*= 1/cweights[j]`), where this would be definitely done once. :)

tlienart:
Apart from the broadcasting questions which are good for another PR (and possible refactoring of the code), is there anything else left to change before a merge?

src/kmeans.jl
```diff
@@ -18,35 +18,34 @@ const _kmeans_default_maxiter = 100
 const _kmeans_default_tol = 1.0e-6
 const _kmeans_default_display = :none
 
-function kmeans!(X::Matrix{T}, centers::Matrix{T};
-                 weights=nothing,
+function kmeans!(X::AbstractMatrix{T}, centers::AbstractMatrix{T};
+                 weights::Union{Nothing, Vector{<:Real}}=nothing,
```
```suggestion
                 weights::Union{Nothing, AbstractVector{<:Real}}=nothing,
```

src/kmeans.jl
```diff
@@ -64,27 +63,27 @@ end
...
+    centers::AbstractMatrix{T},   # in/out: matrix of centers (d x k)
+    assignments::Vector{Int},     # out: vector of assignments (n)
+    costs::Vector{T},             # out: costs of the resultant assignments (n)
+    counts::Vector{Int},          # out: # samples assigned to each cluster (k)
+    cweights::Vector{Float64},    # out: weights of each cluster
+    maxiter::Integer,             # in: maximum number of iterations
+    tol::Real,                    # in: tolerance of change at convergence
+    displevel::Int,               # in: the level of display
+    distance::SemiMetric          # in: function to calculate the distance with
+    ) where T<:Real
```
alyst: The pre-PR code contained both indented and unindented examples of a stray closing bracket + type constraints declaration.
I like unindented version more as it helps to separate the function arguments from its body. @nalimilan What is the recommended convention?

```suggestion
) where T<:Real
```

alyst:
LGTM except `weights` declaration in `kmeans!()`.

src/kmeans.jl
tlienart: to be honest it's not really like the current code (now and before) is super consistent in terms of style... but can we work on this in another PR potentially with broadcasting and type fix?
To be fair, I'm considering rewriting the whole backbone of this `kmeans.jl` file so that it can use more than one algorithm and can take a `Table` instead of just matrices... style questions and better docstrings could go in there too.

nalimilan: I don't think there's a convention about this. In general the Julia code base doesn't use that style, but instead puts the first argument on the same line as the function name.

src/kmeans.jl
nalimilan: I agree, the broadcasted version was nicer. Better use `view`. Same above, where `view` was used but you removed it: use `xj .* wj` and everything should be OK.

src/kmeans.jl
```diff
@@ -18,35 +18,34 @@ const _kmeans_default_maxiter = 100
...
-    assignments = zeros(Int, n)
-    costs = zeros(T, n)
+    assignments = Vector{Int}(undef, n)
+    costs = Vector{T}(undef, n)
     counts = Vector{Int}(undef, k)
     cweights = Vector{Float64}(undef, k)
 
     _kmeans!(X, conv_weights(T, n, weights), centers,
-             assignments, costs, counts, cweights,
-             round(Int, maxiter), tol, display_level(display), distance)
```
nalimilan: This is a typical trick used to avoid recompiling the function for each `maxiter` type. It's probably not very useful here, but I see no reason to remove it either.

nalimilan:
Fine with me, but then please revert all broadcasting changes which aren't related to the main purpose of this PR. It's never a good idea to mix different issues in a single PR (as we can see).

src/kmeans.jl
tlienart: ok

tlienart:
> Fine with me, but then please revert all broadcasting changes which aren't related to the main purpose of this PR. It's never a good idea to mix different issues in a single PR (as we can see).

I don't think there's any non-trivial change in that regards that are left?

Issue #140, opened 2019-01-28
Potential type issues with `fuzzy_cmeans`


tlienart:
Following explorations in #138 I had a look at the code for `fuzzy_cmeans` and tried:

```julia
julia> Random.seed!(51)
julia> X = rand(Int16, 100, 5);
julia> fuzzy_cmeans(X, 3, 1.2)
ERROR: InexactError: Int16(-9526.902803360517)
```

I believe this may stem from the fact that the algorithm accepts any subtype of `Real` and therefore ints but that with ints there may be type issues with how `centers` are assigned and updated. 

Restricting the types for `X` to `AbstractMatrix{T} where T<:AbstractFloat` solves the issue I believe. So maybe we can do that or if being able to deal with Ints is desirable, just convert `X` to `float` if it isn't?

nalimilan:
Solution 1. isn't very interesting since it fixes the error by throwing another error. Better choose the appropriate element type using `float(eltype(X))` or something similar.

tlienart:
I agree. The disadvantage being that it requires converting the matrix to a floating point one, so effectively copying the data. 

With respect to `float(eltype(X))`, one thing I noted is that `float(Int)` for `Int16, Int32, ...` always gives `Float64`. That's not super intuitive to me. Wouldn't it be preferrable to use the a `Float32` if we can to speed up computations?

alyst:
> The disadvantage being that it requires converting the matrix to a floating point one, so effectively copying the data.

Why so? The `centers` are calculated in `update_centers!()`, and from that code it doesn't look like you need to do any explicit conversion of `data` matrix or its elements.

> With respect to float(eltype(X)), one thing I noted is that float(Int) for Int16, Int32, ... always gives Float64. That's not super intuitive to me. Wouldn't it be preferrable to use the a Float32 if we can to speed up computations?

`float(<:Integer)` returns the most reasonable type, which is the one that is most precise.
Note that for `Int16` the maximal number is 32767, but `maxintfloat(Float16)` is 2048.0; not every larger integer could be precisely represented as `Float16`.
But maybe it would make sense -- similar to `kmeans!()` -- to have `fuzzy_cmeans!(data, centers)` method, which would allow specifying the centers eltype.
And speaking of `kmeans!()`: actually after #138 it's still not possible that `centers` have eltype different from `X`.

tlienart:
In `kmeans`, these lines:

```julia
centers[i, cj] += X[i, j]
```

and

[...]

PR #143, opened 2019-01-30
refactoring of kmeans + type fixes

tlienart:
With apologies for being back so soon... 

This applies the following changes

"**major**"
1. `x_int = rand(Int16, 10, 20); kmeans(x_int, 3)` ~~is now processed by converting `x_int`  into `Float64` (i.e. something sensible is returned instead of an error)~~ now works by making sure the centers are of an appropriately wide type see #140 for which a similar solution could be applied. Test added for that case.
2. types are now checked for consistency. Effectively at the entry points, the types for `X`, `centers` and `weights` can be whatever the user wants ~~but `_kmeans!` is effectively called on one single type which is `eltype(X)`. Since this happens after conversion, we're sure the result is a floating point number of some type `T` and all operations, distances and updates are done in this type `T`.~~

"**minor**"
- following comments from #138, the style is unified somewhat with identation etc. 
- broadcasting with views is still not applied (I tried a number of ways, all were slower and allocated more) but some simplifications are applied (e.g. the bound check for `1<=assignment<=k` is always verified + in-place computations where appropriate.

I've added comments preceded by "NOTE REFACT:" to justify some of the changes hoping that would help with the reviews, these would be eventually removed of course.

**Mini benchmark**

(same as for #138)

**before**
```julia
julia> using BenchmarkTools, Random, Clustering; Random.seed!(1641); X = randn(500, 5000); w = abs.(randn(5000)); @btime kmeans($X, 3); @btime kmeans($X, 3, weights=w);
  137.222 ms (278 allocations: 1.63 MiB)
  95.003 ms (184 allocations: 1.13 MiB)
```

**after** (992c8bb)
```julia
julia> using BenchmarkTools, Random, Clustering; Random.seed!(1641); X = randn(500, 5000); w = abs.(randn(5000)); @btime kmeans($X, 3); @btime kmeans($X, 3, weights=w);
  106.599 ms (230 allocations: 1.32 MiB)
  83.018 ms (180 allocations: 1.05 MiB)
```

### DIFF
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -204,7 +204,7 @@ function update_assignments!(dmat::Matrix{T},          # in:  distance matrix (k
         # find the closest cluster to the i-th sample. Note that a
         # is necessarily between 1 and size(dmat, 1) === k as a result
         # and can thus be used as an index in an `inbounds` environment
-        a = argmin(view(dmat, :, j))
+        c, a = findmin(view(dmat, :, j))
         c = dmat[a, j]
 
         # set/update the assignment
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -69,7 +69,7 @@ function kmeans(X::AbstractMatrix{<:Real},                # in: sample matrix (p
                 distance::SemiMetric=SqEuclidean())       # in: function to calculate distance with
     p, n = size(X)
     k = round(Int, k)
-    (2 <= k < n) || error("k must have 2 <= k < n.")
+    (2 <= k < n) || throw(ArgumentError("k must be 2 <= k < n, k=$k given."))
 
     # initialize the centers using a type wide enough so that the updates
     # centers[i, cj] += X[i, j] * wj will occur without loss of precision through rounding
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -48,13 +48,6 @@ function kmeans!(centers::AbstractMatrix{<:AbstractFloat}, # in: current centers
     assignments = Vector{Int}(undef, n)
     counts = Vector{Int}(undef, k)
 
-    # check the types to see wheter the updates: centers[i, cj] += X[i, j] * wj
-    # may occur loss of precision through silent casting
-    update_type = float(weights === nothing ? eltype(X) : typeof(one(eltype(X)) * one(eltype(weights))))
-    update_type <: eltype(centers) || @warn "The type of the centers update ($update_type) is " *
-                                            "wider than that of the given centers " *
-                                            "($(eltype(centers))). This may incur rounding errors."
-
     _kmeans!(X, weights, centers, assignments, counts,
              Int(maxiter), Float64(tol), display_level(display), distance)
 end
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -36,10 +36,10 @@ function kmeans!(X::AbstractMatrix{<:Real},                # in: sample matrix (
                  tol::Real=_kmeans_default_tol,            # in: tolerance of change at convergence
                  display::Symbol=_kmeans_default_display,  # in: level of display
                  distance::SemiMetric=SqEuclidean())       # in: function to compute distances
-    p, n = size(X)
-    p2, k = size(centers)
+    d, n = size(X)
+    dc, k = size(centers)
 
-    p == p2 || throw(DimensionMismatch("Inconsistent array dimensions for `X` and `centers`."))
+    d == dc || throw(DimensionMismatch("Inconsistent array dimensions for `X` and `centers`."))
     (2 <= k < n) || error("k must have 2 <= k < n.")
     if weights !== nothing
         length(weights) == n || throw(DimensionMismatch("Incorrect length of weights."))
@@ -64,7 +64,7 @@ function kmeans(X::AbstractMatrix{<:Real},                # in: sample matrix (d
                 tol::Real=_kmeans_default_tol,            # in: tolerance  of change at convergence
                 display::Symbol=_kmeans_default_display,  # in: level of display
                 distance::SemiMetric=SqEuclidean())       # in: function to calculate distance with
-    p, n = size(X)
+    d, n = size(X)
     k = round(Int, k)
     (2 <= k < n) || throw(ArgumentError("k must be 2 <= k < n, k=$k given."))
 
@@ -72,7 +72,7 @@ function kmeans(X::AbstractMatrix{<:Real},                # in: sample matrix (d
     # centers[i, cj] += X[i, j] * wj will occur without loss of precision through rounding
     T = float(weights === nothing ? eltype(X) : promote_type(eltype(X), eltype(weights)))
     iseeds = initseeds(init, X, k)
-    centers = copyseeds!(Matrix{T}(undef, p, k), X, iseeds)
+    centers = copyseeds!(Matrix{T}(undef, d, k), X, iseeds)
 
     kmeans!(X, centers;
             weights=weights, maxiter=Int(maxiter), tol=Float64(tol),
@@ -89,7 +89,7 @@ function _kmeans!(X::AbstractMatrix{<:Real},                # in: sample matrix 
                   tol::Float64,                             # in: tolerance of change at convergence
                   displevel::Int,                           # in: the level of display
                   distance::SemiMetric)                     # in: function to calculate distance
-    p, n = size(X)
+    d, n = size(X)
     k = size(centers, 2)
     to_update = Vector{Bool}(undef, k) # whether a center needs to be updated
     unused = Vector{Int}()
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -5,7 +5,7 @@
 ####
 
 struct KmeansResult{TC<:AbstractFloat,TD<:Real,TCW<:Real} <: ClusteringResult
-    centers::Matrix{TC}       # cluster centers (d x k)
+    centers::Matrix{TC}       # cluster centers (p x k)
     assignments::Vector{Int}  # assignments (n)
     costs::Vector{TD}         # cost of the assignments (n)
     counts::Vector{Int}       # number of samples assigned to each cluster (k)
@@ -24,12 +24,13 @@ const _kmeans_default_display = :none
 #### Exported kmeans and kmeans! functions
 ####
 
-function kmeans!(X::AbstractMatrix{TX}, centers::AbstractMatrix{TC};
-                 weights::Union{Nothing, AbstractVector{TW}}=nothing,
-                 maxiter::Integer=_kmeans_default_maxiter,
-                 tol::Real=_kmeans_default_tol,
-                 display::Symbol=_kmeans_default_display,
-                 distance::SemiMetric=SqEuclidean()
+function kmeans!(X::AbstractMatrix{TX},                    # in: sample matrix (p x n)
+                 centers::AbstractMatrix{TC};              # in: current centers (p x k)
+                 weights::Union{Nothing, AbstractVector{TW}}=nothing, # in: sample weights (n)
+                 maxiter::Integer=_kmeans_default_maxiter, # in: maximum number of iterations
+                 tol::Real=_kmeans_default_tol,            # in: tolerance of change at convergence
+                 display::Symbol=_kmeans_default_display,  # in: level of display
+                 distance::SemiMetric=SqEuclidean()        # in: function to compute distances
                  ) where {TX<:Real,TC<:Real,TW<:Real}
     p, n = size(X)
     p2, k = size(centers)
@@ -55,13 +56,15 @@ function kmeans!(X::AbstractMatrix{TX}, centers::AbstractMatrix{TC};
              display_level(display), distance)
 end
 
-function kmeans(X::AbstractMatrix{TX}, k::Int;
-                weights::Union{Nothing, AbstractVector{TW}}=nothing,
-                init=_kmeans_default_init,
-                maxiter::Integer=_kmeans_default_maxiter,
-                tol::Real=_kmeans_default_tol,
-                display::Symbol=_kmeans_default_display,
-                distance::SemiMetric=SqEuclidean()
+
+function kmeans(X::AbstractMatrix{TX}, # in: sample matrix (p x n) columns = observations
+                k::Int;                # in: number of centers
+                weights::Union{Nothing, AbstractVector{TW}}=nothing, # in: sample weights (n)
+                init=_kmeans_default_init,                # in: initialization algorithm
+                maxiter::Integer=_kmeans_default_maxiter, # in: maximum number of iterations
+                tol::Real=_kmeans_default_tol,            # in: tolerance  of change at convergence
+                display::Symbol=_kmeans_default_display,  # in: level of display
+                distance::SemiMetric=SqEuclidean()        # in: function to calculate distance with
                 ) where {TX<:Real,TW<:Real}
     p, n = size(X)
     (2 <= k < n) || error("k must have 2 <= k < n.")
@@ -82,15 +85,15 @@ end
 #### Core implementation
 ####
 
-function _kmeans!(X::AbstractMatrix{TX},
-                  weights::Union{Nothing, Vector{TW}},
-                  centers::Matrix{TC},
-                  assignments::Vector{Int},
-                  counts::Vector{Int},
-                  maxiter::Int,
-                  tol::Real,
-                  displevel::Int,
-                  distance::SemiMetric
+function _kmeans!(X::AbstractMatrix{TX},    # in: sample matrix (p x n)
+                  weights::Union{Nothing, Vector{TW}}, # in: sample weights (n)
+                  centers::Matrix{TC},      # in/out: matrix of centers (p x k)
+                  assignments::Vector{Int}, # out: vector of assignments (n)
+                  counts::Vector{Int},      # out: # samples assigned to each cluster (k)
+                  maxiter::Int,             # in: maximum number of iterations
+                  tol::Real,                # in: tolerance of change at convergence
+                  displevel::Int,           # in: the level of display
+                  distance::SemiMetric      # in: function to calculate the distance with
                   ) where {TX<:Real,TW<:Real,TC<:AbstractFloat}
     p, n = size(X)
     k = size(centers, 2)
@@ -175,10 +178,13 @@ end
 #### update assignments
 ####
 
-function update_assignments!(dmat::Matrix{T}, is_init::Bool,
-                             assignments::Vector{Int}, costs::Vector{T},
-                             counts::Vector{Int}, to_update::Vector{Bool},
-                             unused::Vector{Int}
+function update_assignments!(dmat::Matrix{T},          # in:  distance matrix (k x n)
+                             is_init::Bool,            # in:  whether it is the initial run
+                             assignments::Vector{Int}, # out: assignment vector (n)
+                             costs::Vector{T},         # out: costs of the resultant assignment (n)
+                             counts::Vector{Int},      # out: # samples assigned to each cluster (k)
+                             to_update::Vector{Bool},  # out: whether a center needs update (k)
+                             unused::Vector{Int}       # out: list of centers with no samples
                              ) where T<:Real
     k, n = size(dmat)
 
@@ -232,9 +238,12 @@ end
 #### update centers (unweighted case: weights=nothing)
 ####
 
-function update_centers!(X::AbstractMatrix{TX}, weights::Nothing,
-                         assignments::Vector{Int}, to_update::Vector{Bool},
-                         centers::Matrix{TC}, cweights::Vector{Int}
+function update_centers!(X::AbstractMatrix{TX},     # in: sample matrix (p x n)
+                         weights::Nothing,          # in: sample weights
+                         assignments::Vector{Int},  # in: assignments (n)
+                         to_update::Vector{Bool},   # in: whether a center needs update (k)
+                         centers::Matrix{TC},       # out: updated centers (p x k)
+                         cweights::Vector{Int}      # out: updated cluster weights (k)
                          ) where {TX<:Real,TC<:AbstractFloat}
     d, n = size(X)
     k = size(centers, 2)
@@ -275,9 +284,12 @@ end
 #### update centers (weighted case)
 ####
 
-function update_centers!(X::AbstractMatrix{TX}, weights::Vector{TW},
-                         assignments::Vector{Int}, to_update::Vector{Bool},
-                         centers::Matrix{TC}, cweights::Vector{TW}
+function update_centers!(X::AbstractMatrix{TX},     # in: sample matrix (p x n)
+                         weights::Vector{TW},       # in: sample weights (n)
+                         assignments::Vector{Int},  # in: assignments (n)
+                         to_update::Vector{Bool},   # in: whether a center needs update (k)
+                         centers::Matrix{TC},       # out: updated centers (p x k)
+                         cweights::Vector{TW}       # out: updated cluster weights (k)
                          ) where {TX<:Real,TC<:AbstractFloat,TW<:Real}
     d, n = size(X)
     k = size(centers, 2)
@@ -321,9 +333,11 @@ end
 #### Re-pick centers that get no samples assigned to them.
 ####
 
-function repick_unused_centers(X::AbstractMatrix{TX}, costs::Vector{TD},
-                               centers::Matrix{TC}, unused::Vector{Int},
-                               distance::SemiMetric
+function repick_unused_centers(X::AbstractMatrix{TX},  # in: the sample set (p x n)
+                               costs::Vector{TD},      # in: the current assignment costs (n)
+                               centers::Matrix{TC},    # to be updated: the centers (p x k)
+                               unused::Vector{Int},    # in: set of indices of centers to be updated
+                               distance::SemiMetric    # in: function to calculate the distance with
                                ) where {TX<:Real,TC<:AbstractFloat,TD<:Real}
     # pick new centers using a scheme like kmeans++
     ds = similar(costs)
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -1,8 +1,6 @@
 # K-means algorithm
 
-####
-#### Result object
-####
+#### Interface
 
 # T is the eltype(centers)
 # D is the type of pairwise distance computation from samples to centers
@@ -23,10 +21,6 @@ const _kmeans_default_init = :kmpp
 const _kmeans_default_maxiter = 100
 const _kmeans_default_tol = 1.0e-6
 const _kmeans_default_display = :none
-
-####
-#### Exported kmeans and kmeans! functions
-####
 
 """
     kmeans!(centers, X)
@@ -95,11 +89,9 @@ function kmeans(X::AbstractMatrix{<:Real},                # in: sample matrix (p
             display=display, distance=distance)
 end
 
-
-####
 #### Core implementation
-####
 
+# core k-means skeleton
 function _kmeans!(X::AbstractMatrix{<:Real},                # in: sample matrix (p x n)
                   weights::Union{Nothing, Vector{<:Real}},  # in: sample weights (n)
                   centers::AbstractMatrix{<:AbstractFloat}, # in/out: matrix of centers (p x k)
@@ -188,10 +180,10 @@ function _kmeans!(X::AbstractMatrix{<:Real},                # in: sample matrix 
                         cweights, objv, t, converged)
 end
 
-####
-#### update assignments
-####
-
+#
+#  Updates assignments, costs, and counts based on
+#  an updated (squared) distance matrix
+#
 function update_assignments!(dmat::Matrix{T},          # in:  distance matrix (k x n)
                              is_init::Bool,            # in:  whether it is the initial run
                              assignments::Vector{Int}, # out: assignment vector (n)
@@ -250,10 +242,11 @@ function update_assignments!(dmat::Matrix{T},          # in:  distance matrix (k
     end
 end
 
-####
-#### update centers (unweighted case)
-####
-
+#
+#  Update centers based on updated assignments
+#
+#  (specific to the case where samples are not weighted)
+#
 function update_centers!(X::AbstractMatrix{<:Real},        # in: sample matrix (p x n)
                          weights::Nothing,                 # in: sample weights
                          assignments::Vector{Int},         # in: assignments (n)
@@ -295,10 +288,11 @@ function update_centers!(X::AbstractMatrix{<:Real},        # in: sample matrix (
     end
 end
 
-####
-#### update centers (weighted case)
-####
-
+#
+#  Update centers based on updated assignments
+#
+#  (specific to the case where samples are weighted)
+#
 function update_centers!(X::AbstractMatrix{<:Real}, # in: sample matrix (p x n)
                          weights::Vector{W},        # in: sample weights (n)
                          assignments::Vector{Int},  # in: assignments (n)
@@ -344,10 +338,9 @@ function update_centers!(X::AbstractMatrix{<:Real}, # in: sample matrix (p x n)
 end
 
 
-####
-#### Re-pick centers that get no samples assigned to them.
-####
-
+#
+#  Re-picks centers that get no samples assigned to them.
+#
 function repick_unused_centers(X::AbstractMatrix{<:Real},  # in: the sample set (p x n)
                                costs::Vector{<:Real},      # in: the current assignment costs (n)
                                centers::AbstractMatrix{<:AbstractFloat}, # out: the centers (p x k)
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -49,7 +49,7 @@ function kmeans!(X::AbstractMatrix{TX},                    # in: sample matrix (
     TC2 = promote_type(TX, TC)
     weights === nothing || (TC2 = promote_type(TC2, TW))
     # corner case where that's still not an AbstractFloat
-    TC2 = ifelse(TC2 <: AbstractFloat, TC2, Float64)
+    TC2 = float(TC2)
 
     _kmeans!(X, weights, convert(Matrix{TC2}, centers),
              assignments, counts, round(Int, maxiter), tol,
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -20,14 +20,13 @@ const _kmeans_default_maxiter = 100
 const _kmeans_default_tol = 1.0e-6
 const _kmeans_default_display = :none
 
-
 ####
 #### Exported kmeans and kmeans! functions
 ####
 
 function kmeans!(X::AbstractMatrix{TX}, centers::AbstractMatrix{TC};
                  weights::Union{Nothing, AbstractVector{TW}}=nothing,
-                 maxiter::Int=_kmeans_default_maxiter,
+                 maxiter::Integer=_kmeans_default_maxiter,
                  tol::Real=_kmeans_default_tol,
                  display::Symbol=_kmeans_default_display,
                  distance::SemiMetric=SqEuclidean()
@@ -51,19 +50,15 @@ function kmeans!(X::AbstractMatrix{TX}, centers::AbstractMatrix{TC};
     # corner case where that's still not an AbstractFloat
     TC2 = ifelse(TC2 <: AbstractFloat, TC2, Float64)
 
-    # NOTE REFACT:
-    # - X is still of type TX<:Real (no conversion)
-    # - centers is of type T <: AbstractFloat (conversion only in corner case)
-    # - weights is still of type TW <: Real (no conversion)
     _kmeans!(X, weights, convert(Matrix{TC2}, centers),
-             assignments, counts, maxiter, tol,
+             assignments, counts, round(Int, maxiter), tol,
              display_level(display), distance)
 end
 
 function kmeans(X::AbstractMatrix{TX}, k::Int;
                 weights::Union{Nothing, AbstractVector{TW}}=nothing,
                 init=_kmeans_default_init,
-                maxiter::Int=_kmeans_default_maxiter,
+                maxiter::Integer=_kmeans_default_maxiter,
                 tol::Real=_kmeans_default_tol,
                 display::Symbol=_kmeans_default_display,
                 distance::SemiMetric=SqEuclidean()
@@ -71,17 +66,14 @@ function kmeans(X::AbstractMatrix{TX}, k::Int;
     p, n = size(X)
     (2 <= k < n) || error("k must have 2 <= k < n.")
 
-    # NOTE REFACT: if eltype(X) is not AbstractFloat, centers are
-    # initialised as a Float64 conversion of k chosen elements of X
-    # (but X itself is not converted)
-
-    TC = ifelse(TX <: AbstractFloat, TX, Float64)
+    TC = float(TX)
+    centers = Matrix{TC}(undef, p, k)
 
     iseeds = initseeds(init, X, k)
-    centers = copyseeds(X, iseeds, TC)
+    copyseeds!(centers, X, iseeds)
 
     kmeans!(X, centers;
-            weights=weights, maxiter=maxiter, tol=tol,
+            weights=weights, maxiter=round(Int, maxiter), tol=tol,
             display=display, distance=distance)
 end
 
@@ -90,8 +82,6 @@ end
 #### Core implementation
 ####
 
-# NOTE REFACT: this is called after `kmeans!` where X and weights have not
-# been touched but centers has been coerced to be of type AbstractFloat
 function _kmeans!(X::AbstractMatrix{TX},
                   weights::Union{Nothing, Vector{TW}},
                   centers::Matrix{TC},
@@ -108,12 +98,9 @@ function _kmeans!(X::AbstractMatrix{TX},
     unused = Vector{Int}()
     num_affected = k # number of centers to which dists need to be recomputed
 
-    # compute pairwise distances
+    # compute pairwise distances, preassign costs and cluster weights
     dmat = pairwise(distance, centers, X)
-
-    # NOTE REFACT: costs is only associated to dmat hence its type defined here
     costs = Vector{eltype(dmat)}(undef, n)
-    # NOTE REFACT: cluster weights is either incremented by 1 or by weight
     TCW = weights === nothing ? Int : TW
     cweights = Vector{TCW}(undef, k)
 
@@ -168,10 +155,9 @@ function _kmeans!(X::AbstractMatrix{TX},
 
         # display information (if required)
         if displevel >= 2
-            @printf("%7d %18.6e %18.6e | %8d\n", t, objv, objv_change,
-                    num_affected)
+            @printf("%7d %18.6e %18.6e | %8d\n", t, objv, objv_change, num_affected)
         end
-    end # end while
+    end
 
     if displevel >= 1
         if converged
@@ -246,8 +232,6 @@ end
 #### update centers (unweighted case: weights=nothing)
 ####
 
-# NOTE REFACT: we know here that if weights is nothing then cweights must
-# be of type Int.
 function update_centers!(X::AbstractMatrix{TX}, weights::Nothing,
                          assignments::Vector{Int}, to_update::Vector{Bool},
                          centers::Matrix{TC}, cweights::Vector{Int}
@@ -265,9 +249,6 @@ function update_centers!(X::AbstractMatrix{TX}, weights::Nothing,
         if to_update[cj]
             if cweights[cj] > 0
                 for i ∈ 1:d
-                    # NOTE REFACT: centers is of type TC<:AbstractFloat
-                    # which is a supertype of (TX,TC,TW) and so there is
-                    # no loss of precision here. (same below)
                     centers[i, cj] += X[i, j]
                 end
             else
@@ -302,7 +283,7 @@ function update_centers!(X::AbstractMatrix{TX}, weights::Vector{TW},
     k = size(centers, 2)
 
     # initialize center weights
-    cweights[to_update] .= zero(TW)
+    cweights[to_update] .= 0
 
     # accumulate columns
     @inbounds for j ∈ 1:n
@@ -310,7 +291,7 @@ function update_centers!(X::AbstractMatrix{TX}, weights::Vector{TW},
         # that doesn't need to be updated
         wj = weights[j]
         cj = assignments[j]
-        if to_update[cj] && wj > 0
+        if wj > 0 && to_update[cj]
             if cweights[cj] > 0
                 for i ∈ 1:d
                     centers[i, cj] += X[i, j] * wj
@@ -353,7 +334,6 @@ function repick_unused_centers(X::AbstractMatrix{TX}, costs::Vector{TD},
         j = wsample(1:n, tcosts)
         tcosts[j] = 0
         v = view(X, :, j)
-        # NOTE REFACT: centers is of type TC which is a supertype of TX
         copyto!(centers[:, i], v)
         colwise!(ds, distance, v, X)
         tcosts = min(tcosts, ds)
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -4,15 +4,19 @@
 #### Result object
 ####
 
-struct KmeansResult{TC<:AbstractFloat,TD<:Real,TCW<:Real} <: ClusteringResult
-    centers::Matrix{TC}       # cluster centers (p x k)
-    assignments::Vector{Int}  # assignments (n)
-    costs::Vector{TD}         # cost of the assignments (n)
-    counts::Vector{Int}       # number of samples assigned to each cluster (k)
-    cweights::Vector{TCW}     # cluster weights (k)
-    totalcost::TD             # total cost (i.e. objective)
-    iterations::Int           # number of elapsed iterations
-    converged::Bool           # whether the procedure converged
+# T is the eltype(centers)
+# D is the type of pairwise distance computation from samples to centers
+# WC is the type of cluster weights, either Int (in the case where samples are
+# unweighted) or eltype(weights) (in the case where samples are weighted).
+struct KmeansResult{T<:AbstractFloat,D<:Real,WC<:Real} <: ClusteringResult
+    centers::AbstractMatrix{T} # cluster centers (p x k)
+    assignments::Vector{Int}   # assignments (n)
+    costs::Vector{D}           # cost of the assignments (n)
+    counts::Vector{Int}        # number of samples assigned to each cluster (k)
+    cweights::Vector{WC}       # cluster weights (k)
+    totalcost::D               # total cost (i.e. objective)
+    iterations::Int            # number of elapsed iterations
+    converged::Bool            # whether the procedure converged
 end
 
 const _kmeans_default_init = :kmpp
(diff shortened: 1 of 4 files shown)


src/kmeans.jl
```diff
@@ -1,91 +1,132 @@
...
-    totalcost::Float64         # total cost (i.e. objective) (k)
-    iterations::Int            # number of elapsed iterations
-    converged::Bool            # whether the procedure converged
+####
+#### Result object
+####
+
+# NOTE REFACT: there's no need for this to be a mutable struct.
+
+struct KmeansResult{T<:AbstractFloat} <: ClusteringResult
```
alyst: Actually, rather then tightening the constraint, I think it could be relaxed even further to `<: Number` (i.e. complex numbers).
Anyway, the inference and/or checking of `T` would be done by `kmeans[!]()`.

tlienart: Do you have a reference/use-case for kmeans on Complex Number? This could be for another PR though...
Here, I tightened the parameter of that type because it never got anything else.

src/kmeans.jl
```diff
@@ -1,91 +1,132 @@
...
+    centers::Matrix{T}        # cluster centers (d x k)
+    assignments::Vector{Int}  # assignments (n)
+    costs::Vector{T}          # cost of the assignments (n)
```
alyst: I don't see why `costs` and `weights` have to be equal to `T`.
I guess, `eltype(costs)` should be something like `promote_type(typeof(float(abs2(zero(T))))), eltype(weights))`.

tlienart: there are in the PR after conversion (see `kmeans!`). But maybe wait a bit for the full review as I'm implementing further changes to remove more conversions as per comments in #140 .

src/kmeans.jl
```diff
@@ -1,91 +1,132 @@
...
+                 centers::AbstractMatrix{<:Real}; args...)
+    # NOTE REFACT: a warning could be issued here
+    kmeans!(convert(Matrix{Float64}, X), centers; args...)
+end
+
+# case when TX<:AbstractFloat
+function kmeans!(X::AbstractMatrix{TX}, centers::AbstractMatrix{TC};
                  weights::Union{Nothing, AbstractVector{<:Real}}=nothing,
-                 maxiter::Integer=_kmeans_default_maxiter,
+                 maxiter::Int=_kmeans_default_maxiter,
```
alyst: I don't see why `maxiter` has to be `Int`. What's wrong with e.g. `0x32`?

tlienart: I think that's just a gimmick to be honest...

tlienart: (and also that's what's used in `fuzzy_cmeans` for instance)

src/kmeans.jl
```diff
@@ -1,91 +1,132 @@
...
-    m, n = size(X)
-    m2, k = size(centers)
-    m == m2 || throw(DimensionMismatch("Inconsistent array dimensions."))
+    p == p2 || throw(DimensionMismatch("Inconsistent array dimensions " *
+                                       "for `X` and `centers`."))
     (2 <= k < n) || error("k must have 2 <= k < n.")
 
     assignments = Vector{Int}(undef, n)
-    costs = Vector{T}(undef, n)
+    costs = Vector{TX}(undef, n) # distance to centroid, will be of type TX.
```
alyst: See the comment for `KmeansResult`, I think it's possible to calculate what type `costs` should be (and that would be inferred at compile time).

tlienart: Yes, I'm moving this out of this function and into `_kmeans!`, it's not necessary for it to be there and in `_kmeans!` the type is explicit once `dmat` has been instantiated.

src/kmeans.jl
```diff
@@ -1,91 +1,132 @@
...
+# case when TX<:AbstractFloat
+function kmeans!(X::AbstractMatrix{TX}, centers::AbstractMatrix{TC};
                  weights::Union{Nothing, AbstractVector{<:Real}}=nothing,
-                 maxiter::Integer=_kmeans_default_maxiter,
+                 maxiter::Int=_kmeans_default_maxiter,
                  tol::Real=_kmeans_default_tol,
                  display::Symbol=_kmeans_default_display,
-                 distance::SemiMetric=SqEuclidean()) where T<:Real
+                 distance::SemiMetric=SqEuclidean()
+                 ) where {TX<:AbstractFloat,TC<:Real}
```
alyst: Actually, `TC <: AbstractFloat` would make more sense here.
`MethodError` indicating that integral `centers` are not supported would be much clearer than `InexactError` in `update_centers!()`

tlienart: yes this is done in 992c8bb 😛

src/kmeans.jl
```diff
@@ -1,91 +1,125 @@
...
+####
+#### Result object
+####
+
+struct KmeansResult{TC<:AbstractFloat,TD<:Real,TCW<:Real} <: ClusteringResult
+    centers::Matrix{TC}       # cluster centers (d x k)
+    assignments::Vector{Int}  # assignments (n)
+    costs::Vector{TD}         # cost of the assignments (n)
+    counts::Vector{Int}       # number of samples assigned to each cluster (k)
+    cweights::Vector{TCW}     # cluster weights (k)
```
alyst: I'm not sure 100% we need so many type parameters, maybe `Vector{<:Real}` or `Vector{Float64}` would do just fine.
@nalimilan ?
Anyway, I'd use `T` instead of `TC`, `C` instead of `TD`, `W` instead of `TCW`.

tlienart: So the reasoning is that now the code in the current PR has no type casting (no loss of precision) and `costs` and `cweights` are defined by the call:

* `costs` is effectively only defined via `dmat` which is defined as the distance between TX and TC elements --> so it's a parameter, it could be encapsulated by a Float64 but I don't think that's necessary/makes more sense
* `cweights` is effectively only defined by `weights`, if `weights==nothing` then it's a vector of Int, otherwise it's a vector of type `eltype(weights)` --> again a parameter

naming of parameters I really don't care so whatever floats your boat. The advantage with these two letter things is that their use is consistent throughout the code and so possibly less confusing. 🤷‍♂️

src/kmeans.jl
alyst: Well, we started by allowing `AbstractMatrix`, and now we are disallowing innocent integral types that could be converted to `Int` with zero overhead :)
This package was started in 2014. Since then Julia evolved quite a bit, so not all of the code here follows the current idiomatic Julia (including `fuzzy_cmeans()`).
Essentially, my hope is that your PRs would improve this situation. :)

tlienart: I'm fine either way and happy to revert if you think that's better though I can't think of any practical application where one would feed the max number of iterations in another type than `Int`... it's just completely pointless.

src/kmeans.jl
```diff
@@ -1,91 +1,125 @@
...
+    end
 
     assignments = Vector{Int}(undef, n)
-    costs = Vector{T}(undef, n)
     counts = Vector{Int}(undef, k)
-    cweights = Vector{Float64}(undef, k)
 
-    _kmeans!(X, conv_weights(T, n, weights), centers,
-             assignments, costs, counts, cweights, round(Int, maxiter), tol,
+    TC2 = promote_type(TX, TC)
```
nalimilan: I don't think this is the place where `centers` should be converted. It's the user's responsibility to provide it in a correct type since it's an in-place function. Anyway it's quite unlikely that one would be able to use an integer type to store centers. And we already choose an appropriate type in `kmeans` right before calling `kmeans!`, so this seems redundant.

src/kmeans.jl
nalimilan: The question should rather be: what's the point in not accepting any `Integer`?

src/kmeans.jl
```diff
@@ -1,91 +1,125 @@
...
+                distance::SemiMetric=SqEuclidean()
+                ) where {TX<:Real,TW<:Real}
+    p, n = size(X)
     (2 <= k < n) || error("k must have 2 <= k < n.")
+
+    # NOTE REFACT: if eltype(X) is not AbstractFloat, centers are
+    # initialised as a Float64 conversion of k chosen elements of X
+    # (but X itself is not converted)
+
+    TC = ifelse(TX <: AbstractFloat, TX, Float64)
```
nalimilan: Just use `float(TX)`.

src/kmeans.jl
```diff
@@ -116,35 +146,32 @@ function _kmeans!(
         else
             # if only a small subset is affected, only compute for that subset
             affected_inds = findall(to_update)
-            dmat_p = pairwise(distance, centers[:, affected_inds], X)
-            dmat[affected_inds, :] .= dmat_p
+            pairwise!(view(dmat, affected_inds, :), distance,
```
nalimilan: I think it's worth benchmarking this change, as sometimes it can be more efficient to make a copy (since computing distances requires going over the array many times).

src/kmeans.jl
```diff
@@ -116,35 +146,32 @@ function _kmeans!(
...
-        update_assignments!(dmat, false, assignments, costs, counts, to_update, unused)
-
+        update_assignments!(dmat, false, assignments, costs, counts,
+                            to_update, unused)
         num_affected = sum(to_update) + length(unused)
 
         # compute change of objective and determine convergence
         prev_objv = objv
-        objv = w === nothing ? sum(costs) : dot(w, costs)
+        objv = weights === nothing ? sum(costs) : dot(weights, costs)
```
nalimilan: I don't see the point of renaming this. Please avoid unnecessary changes as it makes the PR very hard to review.

src/kmeans.jl
```diff
@@ -116,35 +146,32 @@ function _kmeans!(
...
             converged = true
         end
 
-        # display iteration information (if asked)
-
+        # display information (if required)
         if displevel >= 2
-            @printf("%7d %18.6e %18.6e | %8d\n", t, objv, objv_change, num_affected)
+            @printf("%7d %18.6e %18.6e | %8d\n", t, objv, objv_change,
+                    num_affected)
```
nalimilan: This was fine too.

src/kmeans.jl
```diff
@@ -116,35 +146,32 @@ function _kmeans!(
...
         end
-    end
+    end # end while
```
nalimilan: This is not the common Julia style convention AFAIK.

src/kmeans.jl
```diff
@@ -154,25 +181,19 @@ function _kmeans!(
...
-    dmat::Matrix{T},              # in:  distance matrix (k x n)
-    is_init::Bool,                # in:  whether it is the initial run
-    assignments::Vector{Int},     # out: assignment vector (n)
-    costs::Vector{T},             # out: costs of the resultant assignment (n)
-    counts::Vector{Int},          # out: # samples assigned to each cluster (k)
-    to_update::Vector{Bool},      # out: whether a center needs update (k)
-    unused::Vector{Int}          # out: list of centers with no samples
-    ) where T<:Real
-
+function update_assignments!(dmat::Matrix{T}, is_init::Bool,
```
nalimilan: The old one was more readable. It's too bad to discard detailed comments. Same elsewhere.

src/kmeans.jl
```diff
@@ -269,89 +280,82 @@ function update_centers!(
...
+function update_centers!(X::AbstractMatrix{TX}, weights::Vector{TW},
+                         assignments::Vector{Int}, to_update::Vector{Bool},
+                         centers::Matrix{TC}, cweights::Vector{TW}
+                         ) where {TX<:Real,TC<:AbstractFloat,TW<:Real}
     d, n = size(X)
     k = size(centers, 2)
 
     # initialize center weights
-    cweights[to_update] .= 0.0
+    cweights[to_update] .= zero(TW)
```
nalimilan: This should really make no difference in practice. Anyway better use `0` with is the simplest zero type to reason about for the compiler.

src/kmeans.jl
```diff
@@ -269,89 +280,82 @@ function update_centers!(
...
 
     # accumulate columns
-    @inbounds for j = 1:n
+    @inbounds for j ∈ 1:n
+        # skip samples with negative weights or assigned to a center
+        # that doesn't need to be updated
         wj = weights[j]
-        if wj > 0
-            cj = assignments[j]
-            1 <= cj <= k || error("assignment out of boundary.")
```
nalimilan: Are you sure this never happens?

src/kmeans.jl
```diff
@@ -269,89 +280,82 @@ function update_centers!(
...
-                if cweights[cj] > 0
-                    for i = 1:d
-                        centers[i, cj] += X[i, j] * wj
-                    end
-                else
-                    for i = 1:d
-                        centers[i, cj] = X[i, j] * wj
-                    end
+        cj = assignments[j]
+        if to_update[cj] && wj > 0
```
nalimilan: Put `wj > 0` first to avoid an array lookup when not needed.

src/kmeans.jl
```diff
@@ -269,89 +280,82 @@ function update_centers!(
...
-    for i in unused
+    for i ∈ unused
         j = wsample(1:n, tcosts)
         tcosts[j] = 0
         v = view(X, :, j)
-        centers[:, i] = v
-
-        colwise!(ds, SqEuclidean(), v, X)
+        # NOTE REFACT: centers is of type TC which is a supertype of TX
+        copyto!(centers[:, i], v)
```
nalimilan: This won't work AFAICT since it will modify a copy. What's wrong with the old code?

src/seeding.jl
```diff
@@ -40,22 +40,20 @@ initseeds_by_costs(algname::Symbol, costs::AbstractMatrix{<:Real}, k::Integer) =
...
     for j = 1:k
-        copyto!(view(S,:,j), view(X,:,iseeds[j]))
+        copyto!(view(S, :, j), view(X, :, iseeds[j]))
     end
     return S
 end
 
-copyseeds(X::AbstractMatrix{<:Real}, iseeds::AbstractVector) =
-    copyseeds!(similar(X, size(X, 1), length(iseeds)), X, iseeds)
+copyseeds(X::AbstractMatrix{<:Real}, iseeds::AbstractVector, T::DataType) =
```
nalimilan: A more typical Julian approach to this is not to pass the type as an argument, but to allocate the matrix of the right type in the caller and call `copyseeds!` on it.

src/kmeans.jl
nalimilan: I agree single-letter names are enough. Maybe we could drop at least the parameter for `costs`, and use either the same eltype as `centers`? Since we allocate the matrix manually, it shouldn't be an issue. Or just use `Float64`, which is what `dmat` will use AFAICT since it results from `pairwise`.

src/kmeans.jl
tlienart: 🤷‍♂️ as I said, happy to revert.

src/kmeans.jl
tlienart: I don't agree with this, `costs` is exclusively tied to `dmat` which is not guaranteed be of type `centers` and the conversion that existed before to guarantee this is unnecessary IMO.

And re single letter names, it's fine, however there's a different type for `weights` an `clusterweights` and for `centers` and `costs` so that's going to be something like `W`, `E`, `C`, `K` which ends up being less clear imo.

src/kmeans.jl
tlienart: No so `kmeans!` is exposed to the user  so can potentially take anything, indeed in the path where it's `kmeans` that's called then `kmeans!` this promotion doesn't happen.

src/kmeans.jl
tlienart: yeah that's silly, thanks

src/kmeans.jl
tlienart: see small benchmarks at the top, this definitely made a positive change towards fewer allocations and slightly faster overall.

src/kmeans.jl
tlienart: Yeah I don't want to be that guy, however I did try to pick consistent names throughout so that there's no function that calls weights at the top (`kmeans` and `kmeans!`) which then goes on to be called  `w`. I think it'll make further iteration of improvements easier if we stick to the same names for the same things?

src/kmeans.jl
tlienart: ok

src/kmeans.jl
tlienart: yes, `assignments[j]` is set via the `argmin`  call which has for range something that is necessarily between `1` and `k` (otherwise one of the function calls throws an error earlier).

src/kmeans.jl
tlienart: good point

src/kmeans.jl
tlienart: this line was hit in my local tests worked fine and seem to decrease allocations but happy to revert if it's unclear.

src/seeding.jl
tlienart: ok good idea

src/kmeans.jl
tlienart: I'll remove that along with the refactoring notes

src/kmeans.jl
tlienart: ok

src/kmeans.jl
tlienart: actually I tried changing just that back to what it was:

```julia
126.114 ms (280 allocations: 1.63 MiB)
75.811 ms (183 allocations: 1.13 MiB)
```

so it is worse in terms of allocations (timings are not useful here bc I'm another computer).

tlienart:
small bump on this, what do you guys think?

src/kmeans.jl
alyst: I second @nalimilan : it's a very strong Julian convention that `!`-functions modify its corresponding argument in-place (actually, `centers` should be the 1st argument of `kmeans!()`, but I don't know what transition strategy we should use to adopt this API change).
The users are writing code with this convention in mind and violating it will break the their code (e.g. if the users ignore the return value of `kmeans!()` and instead expect the `centers` to be updated).
If it's not possible to modify the argument in-place, the function should fail. In that respect we should restrict `TC` to `TC <: AbstractFloat`: that would also automatically solve `InexactError` issue by throwing much more meaningful `MethodError` beforehand.

src/kmeans.jl
```diff
@@ -230,37 +234,34 @@ function update_assignments!(
...
     d, n = size(X)
     k = size(centers, 2)
 
     # initialize center weights
-    cweights[to_update] .= 0.0
+    cweights[to_update] .= 0
 
     # accumulate columns
-    @inbounds for j = 1:n
+    @inbounds for j ∈ 1:n
```
alyst: Is `∈` mentioned somewhere as recommended style? FWIW, I prefer `in` over `=`, but `∈` unnecessarily introduces Unicode requirement.

src/kmeans.jl
alyst: `copyto!(centers[:, i]) = v`: `centers[:, i]` creates a new array, `copyto!()` modifies it, and then it is goes straight to GC, since nobody references it.
`centers[:, i] = v` is lowered by the compiler into `setindex!(centers, :, i) = v`, which seems right to me; no extra allocation. What exactly was the problem with this line?
(`SqEuclidean` is a nice catch :))

src/kmeans.jl
alyst: It's worth to add a comment about that on `cj = assignments[j]` line: we're in `@inbounds` mode, so OOB indexing may crash Julia.

src/seeding.jl
```diff
@@ -40,22 +40,20 @@ initseeds_by_costs(algname::Symbol, costs::AbstractMatrix{<:Real}, k::Integer) =
...
     for j = 1:k
-        copyto!(view(S,:,j), view(X,:,iseeds[j]))
+        copyto!(view(S, :, j), view(X, :, iseeds[j]))
     end
     return S
 end
 
 copyseeds(X::AbstractMatrix{<:Real}, iseeds::AbstractVector) =
-    copyseeds!(similar(X, size(X, 1), length(iseeds)), X, iseeds)
+    copyseeds!(Matrix{eltype(X)}(undef, size(X, 1), length(iseeds)), X, iseeds)
```
alyst: That makes sense if `X` is `SparseArray`. Is `SparseArray` case covered by any tests?

src/kmeans.jl
```diff
@@ -1,91 +1,115 @@
...
+                ) where {TX<:Real,TW<:Real}
+    p, n = size(X)
     (2 <= k < n) || error("k must have 2 <= k < n.")
+
+    TC = float(TX)
+    centers = Matrix{TC}(undef, p, k)
+
     iseeds = initseeds(init, X, k)
-    centers = copyseeds(X, iseeds)
+    copyseeds!(centers, X, iseeds)
```
alyst: Just a small style suggestion:
```suggestion
    centers = copyseeds!(Matrix{TC}(undef, p, k), X, iseeds)
```
would avoid uninitialized `centers` hanging around.

src/seeding.jl
```diff
@@ -68,7 +66,7 @@ end
 #   choose an arbitrary subset as seeds
 #
 
-mutable struct RandSeedAlg <: SeedingAlgorithm end
+struct RandSeedAlg <: SeedingAlgorithm end
```
alyst: Could you please also convert the comments above into the docstring for this and the other seeding algorithms?

src/kmeans.jl
```diff
@@ -1,91 +1,115 @@
...
-function kmeans(X::AbstractMatrix{<:Real}, k::Int;
-                weights=nothing, init=_kmeans_default_init,
-                maxiter::Integer=_kmeans_default_maxiter,
-                tol::Real=_kmeans_default_tol,
-                display::Symbol=_kmeans_default_display,
-                distance::SemiMetric=SqEuclidean())
 
-    m, n = size(X)
+function kmeans(X::AbstractMatrix{TX}, # in: sample matrix (p x n) columns = observations
+                k::Int;                # in: number of centers
```
```suggestion
                k::Integer;            # in: number of centers
```
Just to clarify the arguments situation. It's good to be nice to the users and relax the type requirements of API function arguments (positional and keyword) as far as possible/reasonable.
However, one have to take into account that Julia would specialize (generate a new compiled version of the method) for each new combination of arguments.
It should be fine for wrapper functions like `kmeans()`, but we should avoid unnecessary specialization on method parameters for the "actual" functions like `_kmeans!()` (e.g. `_kmeans!(..., convert(Int, maxiter), convert(K, tol))`).
Even better, we can inform Julia that we don't need to specialize on certain args of API functions (`@nospecialize(maxiter::Integer=_kmeans_default_maxiter)`) to avoid generating multiple wrappers. @nalimilan Pls correct me if I'm missing something

src/kmeans.jl
tlienart: I'm not convinced that was the point here (the fact that centers gets mutated). 

The point is that the function is exported and therefore a matrix of centers *could* be provided in any format. I don't really see why we would accept anything for `X` but not for `centers` especially when there's virtually no cost incurred by this. 

Note also that `TC <: AbstractFloat` is **not enough** because the centers can be updated by `X_ij * w_j` which could be of a wider type. This is why I suggested this type promotion and relaxed the type to `Real` as the restriction to `AbstractFloat` didn't seem necessary anymore.

Here's a slightly silly example: a user has centers from another code that's somehow in `Float32`, the data `X` and the weights are read as `Float64`, should the updates to centers be downgraded to `Float32` or should the centers be converted to `Float64`?

src/kmeans.jl
tlienart: ok to add a comment

src/kmeans.jl
tlienart: ok fair enough will revert, thanks!

src/kmeans.jl
tlienart: ok!

src/seeding.jl
tlienart: I'll check and will add relevant comment or remove as appropriate

src/kmeans.jl
```diff
@@ -1,91 +1,115 @@
...
+####
+
+function kmeans!(X::AbstractMatrix{TX},                    # in: sample matrix (p x n)
+                 centers::AbstractMatrix{TC};              # in: current centers (p x k)
+                 weights::Union{Nothing, AbstractVector{TW}}=nothing, # in: sample weights (n)
+                 maxiter::Integer=_kmeans_default_maxiter, # in: maximum number of iterations
+                 tol::Real=_kmeans_default_tol,            # in: tolerance of change at convergence
+                 display::Symbol=_kmeans_default_display,  # in: level of display
+                 distance::SemiMetric=SqEuclidean()        # in: function to compute distances
+                 ) where {TX<:Real,TC<:Real,TW<:Real}
```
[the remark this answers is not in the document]
tlienart: Ok for `TX->T` but there needs to be a reference for the type of `weights`  otherwise there will be things like  `eltype(weights)`  floating around and surely that's less efficient than dispatching over it?

src/kmeans.jl
```diff
@@ -1,91 +1,115 @@
...
-                maxiter::Integer=_kmeans_default_maxiter,
-                tol::Real=_kmeans_default_tol,
-                display::Symbol=_kmeans_default_display,
-                distance::SemiMetric=SqEuclidean())
 
-    m, n = size(X)
+function kmeans(X::AbstractMatrix{TX}, # in: sample matrix (p x n) columns = observations
+                k::Int;                # in: number of centers
+                weights::Union{Nothing, AbstractVector{TW}}=nothing, # in: sample weights (n)
+                init=_kmeans_default_init,                # in: initialization algorithm
```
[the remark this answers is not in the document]
tlienart: Ok

src/kmeans.jl
tlienart: ok, I'll look at this with your other comment of maybe removing `copyseeds` 🙂

src/kmeans.jl
alyst: 1. Do we agree that `kmeans!()` should only mutate `centers` array provided by the user, unless we want to break the `!`-functions convention and, most likely, the existing code that uses `kmeans!()`?

2. I see these alternatives for what is allowed for centers:
   1) `centers::AbstractMatrix{<:Real}`, calculated as before. If the result of calculation doesn't fit the eltype of `centers` we get `InexactError` exception, which would be confusing for most users
    2) `centers::AbstractMatrix{<:AbstractFloat}`, calculated as before. That will raise `MethodError` for integer `centers`, which is more informative than `InexactError`. (Note that you can assign integer to float, so `centers[i, k] = X[i,j] * w[i,j]` for integer `X` and `w` should be fine). There still could be cases of `InexactError` for certain combinations of `data`, `weights` and `centers` eltypes, but most frequent ones should be handled by the type constraint. If the user needs another eltype for `centers`, he/she should do the conversion after `kmeans!()` call.
    3) `centers::AbstractMatrix{<:Real}`, but we convert the `X[i,j] * w[i,j]` to fit the eltype of `centers` (e.g. `round(Int, ...)`). The drawback is that it's done implicitly, so the users may not realize that these are not the exact centers. Plus this conversion would generate some overhead. @nalimilan Is there any elegant generic way of doing this conversion?

src/kmeans.jl
alyst: I would be very happy with `centers::Any`, but, as you see, it leads to various issues :)
OTOH, `maxiter::Integer` could be just converted to `Int` without overhead or errors.
Yes, we can leave out `@nospecialize` for now.

src/kmeans.jl
tlienart: Maybe one way to solve both problems would be to deprecate the use of `kmeans!` which doesn't mutate its first argument anyway and maybe suggest a keyword `centers=...` to the `kmeans()` function where that would be taken as the initial centers to start from.

Then there would be no ambiguity with the fact that we convert the `centers` to have the wider type (effectively your iii.) and there would also be no clash with Julia convention with the `!`?

src/kmeans.jl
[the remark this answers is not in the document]
tlienart: Ok so summarising you're in favour of 

1. using `T` for `X` and it being the only explicit parametric type
1. using `eltype(...)` for other things where needed
1. in the result struct, have it be parametrised on a single type (which will be that of centers, not that of `X`) and have the rest (like costs) be declared as `Vector{<:Real}`?

src/kmeans.jl
tlienart: `cweights` is never `nothing`. It's either `Int` (unweighted case) or `eltype(weights)` provided `weights != nothing`.

The use-case is just that if all computations are done in `Float32` it seems silly to widen the type to `Float64`. I actually think the current type parametrization is less error prone than the previous.

Ps: I did add tests for these types.

tlienart:
Ok following comments (thanks for the detailed feedback) here are some modifications (a8d41e4) hopefully going in the right direction.

**major-ish**

1. as per @alyst's comment the convention in Julia with `fun_name!(...)` is not respected by `kmeans!` (mutates second argument). I changed that as it's exported but left the internal functions (that also don't respect the convention but mutate several arguments anyway) untouched.
1. types, I believe we may slowly be moving towards a consensus. In `kmeans!` centers are now forced to be `AbstractFloat` (as per @alyst's suggestion), a warning is raised if the type is not wide enough. In most situations this would not happen as the centers will have generated in `kmeans(...)` and made to have a precisely wide enough type.
1. type naming is simpler in `KmeansResult` as per comments, the only two letter one is `WC` so that it's not mixed with the type of `weights`. 
1. parametrisation of types is implicit in functions unless there are explicitly two arguments 
that must have the same type

**minors**
1. `k::Integer` is now allowed and rounded before the call of `_kmeans...`
1. `init` in `kmeans` is explicitly typed to `::Symbol`
1. `∈` replaced by `in`
1. comment added to specify that `a = argmin(view(....))` is necessarily between `1` and `k` and therefore ok as an index in an `@inbounds` environment
1. basic docstrings added to `kmeans!` and `kmeans`

Ps: apologies for the noob mistake of trying to change too many changes in one PR resulting in big reviews...

src/kmeans.jl
nalimilan: `@nospecialize` would mean `maxiter` would be inferred as `Any` AFAIK. That could be very bad for performance. I wouldn't worry too much about recompilation here, people are not very likely to use custom integer types and it's not the end of the world if some recompilation happens in that case (though converting to `Int` is a good idea if the wrapper already exists).

src/kmeans.jl
```diff
@@ -18,74 +24,106 @@ const _kmeans_default_maxiter = 100
...
+                 maxiter::Integer=_kmeans_default_maxiter, # in: maximum number of iterations
+                 tol::Real=_kmeans_default_tol,            # in: tolerance of change at convergence
+                 display::Symbol=_kmeans_default_display,  # in: level of display
+                 distance::SemiMetric=SqEuclidean())       # in: function to compute distances
+    p, n = size(X)
+    p2, k = size(centers)
+
+    p == p2 || throw(DimensionMismatch("Inconsistent array dimensions for `X` and `centers`."))
     (2 <= k < n) || error("k must have 2 <= k < n.")
+    if !(weights === nothing)
```
```suggestion
    if weights !== nothing
```

src/kmeans.jl
```diff
@@ -18,74 +24,106 @@ const _kmeans_default_maxiter = 100
...
     counts = Vector{Int}(undef, k)
-    cweights = Vector{Float64}(undef, k)
 
-    _kmeans!(X, conv_weights(T, n, weights), centers,
-             assignments, costs, counts, cweights, round(Int, maxiter), tol,
-             display_level(display), distance)
+    # check the types to see wheter the updates: centers[i, cj] += X[i, j] * wj
+    # may occur loss of precision through silent casting
+    update_type = float(weights === nothing ? eltype(X) : promote_type(eltype(X), eltype(weights)))
+    update_type <: eltype(centers) || @warn "The type of the centers update ($update_type) is " *
```
nalimilan: I'm not aware of any functions that do this. `promote_type` is supposed to return a reasonable type, and `<:` doesn't mean anything in terms of precision, so better just assume the type is OK.

Another way to choose the appropriate type would be `typeof(one(eltype(X)) * one(eltype(weights))`. Then you're certain `X[i, j] * wj` will return an element of the same type as the array.

src/kmeans.jl
```diff
@@ -18,74 +24,106 @@ const _kmeans_default_maxiter = 100
...
+                                            "wider than that of the given centers " *
+                                            "($(eltype(centers))). This may incur rounding errors."
+
+    _kmeans!(X, weights, centers, assignments, counts,
+             round(Int, maxiter), tol, display_level(display), distance)
```
```suggestion
             Int(maxiter), tol, display_level(display), distance)
```
(If the caller passes 1.5 we should really throw an error.)

src/kmeans.jl
```diff
@@ -18,74 +24,106 @@ const _kmeans_default_maxiter = 100
...
-
-    # initialize
+####
+#### Core implementation
+####
+
+function _kmeans!(X::AbstractMatrix{<:Real},                # in: sample matrix (p x n)
+                  weights::Union{Nothing, Vector{<:Real}},  # in: sample weights (n)
+                  centers::AbstractMatrix{<:AbstractFloat}, # in/out: matrix of centers (p x k)
+                  assignments::Vector{Int}, # out: vector of assignments (n)
```
nalimilan: Keep alignment.

src/kmeans.jl
```diff
@@ -269,77 +285,74 @@ function update_centers!(
...
-    X::AbstractMatrix{T},            # in: sample matrix (d x n)
-    weights::AbstractVector{<:Real}, # in: sample weights (n)
-    assignments::Vector{Int},        # in: assignments (n)
-    to_update::Vector{Bool},         # in: whether a center needs update (k)
-    centers::AbstractMatrix{T},      # out: updated centers (d x k)
-    cweights::Vector{Float64}        # out: updated cluster weights (k)
-    ) where T<:Real
-
+####
+#### update centers (weighted case: eltype(cweights) == eltype(weights) <: Real)
```
nalimilan: `eltype(cweights) == eltype(weights) <: Real` is redundant with the signature.

src/kmeans.jl
```diff
@@ -1,14 +1,20 @@
 # K-means algorithm
 
-#### Interface
-
-mutable struct KmeansResult{T<:Real} <: ClusteringResult
-    centers::Matrix{T}         # cluster centers (d x k)
+####
```
nalimilan: Do you really need to change this style?

src/kmeans.jl
```diff
@@ -349,9 +362,8 @@ function repick_unused_centers(
         j = wsample(1:n, tcosts)
         tcosts[j] = 0
         v = view(X, :, j)
-        centers[:, i] = v
-
-        colwise!(ds, SqEuclidean(), v, X)
+        centers[:, i] .= v
```
nalimilan: Is `.=` really needed? If not, better use `=`, which is stricter.

src/seeding.jl
nalimilan: What was the problem with `similar`?

nalimilan:
Note that the docs contain a copy of the `struct` definition which also needs updating.

src/kmeans.jl
alyst: Also, to reduce the specializations of `_kmeans()`
```suggestion
             round(Int, maxiter), Float64(tol), display_level(display), distance)
```
I think `Float64` should be fine, the comparison is done once per iteration.

src/seeding.jl
alyst: AFAIK, `similar()` will return a sparse matrix if `X` is sparse. Which is not we want.

alyst:
> Note that the docs contain a copy of the `struct` definition which also needs updating.

Actually, what is the recommended way to organize Julian package documentation nowadays?
It seems a bit suboptimal to document functions and types in the docs, when we can just use the docstrings in-place.
Should we migrate API documentation to docstrings and keep the docs for higher-level things?

nalimilan:
> Actually, what is the recommended way to organize Julian package documentation nowadays?
> It seems a bit suboptimal to document functions and types in the docs, when we can just use the docstrings in-place.
> Should we migrate API documentation to docstrings and keep the docs for higher-level things?

Yes. Ideally docs would be ported from RST to Documenter, and docstrings automatically inserted into the manual.

src/kmeans.jl
alyst: > `cweights` is never `nothing`. It's either `Int` (unweighted case) or `eltype(weights)` provided `weights != nothin`

But for the weightless case shouldn't the `counts` be identical to `cweights`? For me it seems reasonable that if the user did not provide `weights` then `cweights` are also `nothing`.

src/kmeans.jl
tlienart: While I'm happy to use the suggested `typeof(one(eltype(X)) * one(eltype(weights))`, there is still  the question of whether that type is wider than that of centers in `kmeans!`. 

There could be `centers` provided in `Float32` and the result of `typeof(one(eltype(X)) * one(eltype(weights)) ==  Float64`. The resulting update will throw an error: 

```
centers = rand(Float32, 2, 2)
centers[1, 1]  = randn() # error
```

What do you suggest? that this remains uncaught until the actual update which would leave to a hard-to-interpret  stack trace? that we error early ? or that we raise a warning like what  I tried to do here?

src/kmeans.jl
tlienart: alternative is 

```julia
struct KmeansResult{T<:AbstractFloat} <: ClusteringResult
    centers::AbstractMatrix{T} # cluster centers (p x k)
    assignments::Vector{Int}   # assignments (n)
    costs::Vector{<:Real}           # cost of the assignments (n)
    counts::Vector{Int}        # number of samples assigned to each cluster (k)
    cweights::Vector{<:Real}       # cluster weights (k)
    totalcost::Real               # total cost (i.e. objective)
    iterations::Int            # number of elapsed iterations
    converged::Bool            # whether the procedure converged
end
```

I think the parametrized type is cleaner (and makes explicit that costs and totalcost have the same type). I don't think the  previous situation makes sense (casting everything to Float64). So I'm in favour of the change. So I guess my preferences are:

1. as in current suggestion
2. as alternative above
3. as before

src/kmeans.jl
tlienart: ok

tlienart:
Thanks a lot for the reviews. It seems the two remaining sticking points are:

1. in `kmeans!` whether to check for the type of the update to match that of given `centers`  and what to do if it doesn't match (I suggest warning, or error)
2. the result `struct`, whether to parametrize `cost` and `cweight` or not, if not I suggest at least not casting things to Float64 unnecessarily.

Small PS: happy to modify the docs following recommendations.

src/kmeans.jl
nalimilan: That's not an error at all:
```julia
julia> centers = rand(Float32, 2, 2)
2×2 Array{Float32,2}:
 0.821542  0.58411 
 0.304231  0.636419

julia> centers[1, 1]  = randn() # error
-1.9371986249569961

julia> centers[1, 1]
-1.9371986f0
```

src/kmeans.jl
nalimilan: I was talking about the `###`. This introduces changes all over the file, even in places which wouldn't change otherwise.

src/kmeans.jl
tlienart: hmm weird you're right, I'm not sure what I saw... anyway even without error I'd say the question still holds?

src/kmeans.jl
tlienart: Oh! right sorry ok, reverting to initial style.

src/kmeans.jl
nalimilan: Well either you take the eltype of `centers` and assume that's what the user intended, or you allocate a vector with eltype `typeof(one(eltype(X)) * one(eltype(weights))`. But honestly in practice I'm not sure it will make a difference.

src/kmeans.jl
tlienart: done

src/kmeans.jl
tlienart: Ok I removed the weird type check/warning, I agree that at some point we have to assume the user knows that if they feed Float16 centers things might go badly... (and it's unlikely to happen anyway as you pointed out).

tlienart:
Bump? I believe I've either removed or modified all contentious bits of code :)

test/kmeans.jl
```diff
@@ -114,4 +114,13 @@ end
     @test equal_kmresults(r, r_t)
 end
 
+x_int = rand(Int16, m, n)
```
alyst: I'd rename it to `x` and put into "Integer data" testset. Also, we can use `@testset "$T data" for (T, TC, D, WC) in ((Int, Float64, Float64, Int), ...)` to test for more type combinations.

Actually, I don't see that `kmeans!()` is tested anywhere. Could you please also add some tests for it? E.g. would `Float32` centers work with `Float64` data? with `Int` data? Would `Int` centers fail? This was the major motivation of your PRs, so we need to test that.

src/kmeans.jl
```diff
@@ -2,13 +2,17 @@
...
 #### Interface
 
-mutable struct KmeansResult{T<:Real} <: ClusteringResult
-    centers::Matrix{T}         # cluster centers (d x k)
+# T is the eltype(centers)
+# D is the type of pairwise distance computation from samples to centers
+# WC is the type of cluster weights, either Int (in the case where samples are
+# unweighted) or eltype(weights) (in the case where samples are weighted).
+struct KmeansResult{T<:AbstractFloat,D<:Real,WC<:Real} <: ClusteringResult
+    centers::AbstractMatrix{T} # cluster centers (p x k)
```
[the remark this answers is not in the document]
tlienart: I'm against this change as there is still the possibility that someone would feed their centers computed through some other approach and have to transpose it.

Ps: in benchmarks this makes zero significant difference as far as I can tell

src/kmeans.jl
```diff
@@ -18,74 +22,93 @@ const _kmeans_default_maxiter = 100
...
-                 display::Symbol=_kmeans_default_display,
-                 distance::SemiMetric=SqEuclidean()) where T<:Real
-
-    m, n = size(X)
-    m2, k = size(centers)
-    m == m2 || throw(DimensionMismatch("Inconsistent array dimensions."))
+"""
+    kmeans!(centers, X)
+
+Update the current centers `centers` (of size `p x k` where `p` is the dimension and `k` the
```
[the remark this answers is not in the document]
tlienart: it's a convention in most textbook of stats/ml that I know of but I don't really care if `d` is preferred I'll change it back to `d`.

src/kmeans.jl
```diff
@@ -18,74 +22,93 @@ const _kmeans_default_maxiter = 100
...
+number of centroids) using the samples contained in `X` (of size `p x n` where `n` is the number
+of samples).
+"""
+function kmeans!(centers::AbstractMatrix{<:AbstractFloat}, # in: current centers (p x k)
```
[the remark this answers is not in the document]
tlienart: I did this following one of your earlier comment about `kmeans!` not mutating its first argument, but am fine to revert

src/kmeans.jl
```diff
@@ -18,74 +22,93 @@ const _kmeans_default_maxiter = 100
...
+    p, n = size(X)
+    p2, k = size(centers)
+
+    p == p2 || throw(DimensionMismatch("Inconsistent array dimensions for `X` and `centers`."))
     (2 <= k < n) || error("k must have 2 <= k < n.")
+    if weights !== nothing
+        length(weights) == n || throw(DimensionMismatch("Incorrect length of weights."))
+    end
 
     assignments = Vector{Int}(undef, n)
```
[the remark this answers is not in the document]
tlienart: seems reasonable, will do

src/kmeans.jl
```diff
@@ -18,74 +22,93 @@ const _kmeans_default_maxiter = 100
...
+function kmeans(X::AbstractMatrix{<:Real},                # in: sample matrix (p x n) columns = obs
+                k::Integer;                               # in: number of centers
+                weights::Union{Nothing, AbstractVector{<:Real}}=nothing, # in: sample weights (n)
+                init::Symbol=_kmeans_default_init,        # in: initialization algorithm
+                maxiter::Integer=_kmeans_default_maxiter, # in: maximum number of iterations
+                tol::Real=_kmeans_default_tol,            # in: tolerance  of change at convergence
+                display::Symbol=_kmeans_default_display,  # in: level of display
+                distance::SemiMetric=SqEuclidean())       # in: function to calculate distance with
+    p, n = size(X)
+    k = round(Int, k)
```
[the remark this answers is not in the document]
tlienart: yeah and I should use `Int(k)` anyway.

src/kmeans.jl
```diff
@@ -96,18 +119,14 @@ function _kmeans!(
         @printf("%7d %18.6e\n", t, objv)
     end
 
-    while !converged && t < maxiter
-        t = t + 1
+    while !converged && t<maxiter
```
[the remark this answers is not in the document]
tlienart: there's no official style guide :shrug: I personally prefer it when bool conditions are clearly separated from the logical operators but I don't really care, will revert

src/kmeans.jl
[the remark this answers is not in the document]
tlienart: done

src/kmeans.jl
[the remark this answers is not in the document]
alyst: 1) If you believe non-`Matrix` is an important usecase, then, as I said before, you'll have to a) parameterize `KmeansResult` on it (`KMeansResult{MC<:AbstractMatrix{<:Real}...}`); b) add a test case. We should avoid extending API for hypothetical usecases without actually testing them.
2) Sure, it will not affect `res = kmeans()` timings. But it will slow down the [user] code that extensively uses `res.centers`.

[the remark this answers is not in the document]
tlienart: I'm going to be a bit stubborn here and need more convincing to change this. Users of kmeans will mostly want the algorithm to go as quickly as possible and then plug in the centers in some extended feature matrix or will just plot the results, I don't think there's a practical situation where the overhead mentioned in (2) would really be an issue. However I could definitely see someone trying to use `kmeans!(X', centers')`.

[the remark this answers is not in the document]
alyst: Ok, just a toy example showing how this change will affect the type inference of the same function:
```
julia> struct A{T}
       a::AbstractMatrix{T}
       end

julia> struct B{M<:AbstractMatrix{<:Real}}
       a::M
       end

julia> function f(x)
       y = fill!(similar(x.a, size(x.a, 1)), 0)
       @inbounds for j in size(x.a, 2), i in size(x.a, 1)
           y[i] += x.a[i, j]
       end
       return y
       end
f (generic function with 1 method)

julia> @code_warntype f(A(rand(10, 10))
Body::Any
1 ─ %1  = (Base.getfield)(x, :a)::AbstractArray{Float64,2}
│   %2  = (Base.getfield)(x, :a)::AbstractArray{Float64,2}
│   %3  = (Main.size)(%2, 1)::Any
│   %4  = (Main.similar)(%1, %3)::Any
│   %5  = (Main.fill!)(%4, 0)::Any
│   %6  = (Base.getfield)(x, :a)::AbstractArray{Float64,2}
│   %7  = (Main.size)(%6, 2)::Any
│   %8  = (Base.iterate)(%7)::Any
│   %9  = (%8 === nothing)::Bool
│   %10 = (Base.not_int)(%9)::Bool
└──       goto #7 if not %10
2 ┄ %12 = φ (#1 => %8, #6 => %34)::Any
│   %13 = (Core.getfield)(%12, 1)::Any
│   %14 = (Core.getfield)(%12, 2)::Any
│   %15 = (Base.getfield)(x, :a)::AbstractArray{Float64,2}
│   %16 = (Main.size)(%15, 1)::Any
│   %17 = (Base.iterate)(%16)::Any
│   %18 = (%17 === nothing)::Bool
│   %19 = (Base.not_int)(%18)::Bool
└──       goto #5 if not %19
3 ┄ %21 = φ (#2 => %17, #4 => %29)::Any
│   %22 = (Core.getfield)(%21, 1)::Any
│   %23 = (Core.getfield)(%21, 2)::Any
│   %24 = (Base.getindex)(%5, %22)::Any
│   %25 = (Base.getfield)(x, :a)::AbstractArray{Float64,2}
│   %26 = (Base.getindex)(%25, %22, %13)::Any
│   %27 = (%24 + %26)::Any
│         (Base.setindex!)(%5, %27, %22)
│   %29 = (Base.iterate)(%16, %23)::Any
│   %30 = (%29 === nothing)::Bool
│   %31 = (Base.not_int)(%30)::Bool
└──       goto #5 if not %31
4 ─       goto #3
5 ┄ %34 = (Base.iterate)(%7, %14)::Any
│   %35 = (%34 === nothing)::Bool
│   %36 = (Base.not_int)(%35)::Bool
└──       goto #7 if not %36
6 ─       goto #2
7 ┄       return %5

julia> @code_warntype f(B(rand(10, 10))
Body::Array{Float64,1}
1 ─ %1  = (Base.getfield)(x, :a)::Array{Float64,2}
│   %2  = (Base.arraysize)(%1, 1)::Int64
│   %3  = $(Expr(:foreigncall, :(:jl_alloc_array_1d), Array{Float64,1}, svec(Any, Int64), :(:ccall), 2, Array{Float64,1}, :(%2), :(%2)))::Array{Float64,1}
│   %4  = invoke Main.fill!(%3::Array{Float64,1}, 0::Int64)::Array{Float64,1}
│   %5  = (Base.getfield)(x, :a)::Array{Float64,2}
│   %6  = (Base.arraysize)(%5, 2)::Int64
└──       goto #7 if not true
2 ─ %8  = (Base.getfield)(x, :a)::Array{Float64,2}
│   %9  = (Base.arraysize)(%8, 1)::Int64
└──       goto #5 if not true
3 ─ %11 = (Base.arrayref)(false, %4, %9)::Float64
│   %12 = (Base.getfield)(x, :a)::Array{Float64,2}
│   %13 = (Base.arrayref)(false, %12, %9, %6)::Float64
│   %14 = (Base.add_float)(%11, %13)::Float64
│         (Base.arrayset)(false, %4, %14, %9)
└──       goto #5 if not false
4 ─       nothing
5 ┄       goto #7 if not false
6 ─       nothing
7 ┄       return %4
```

test/kmeans.jl
tlienart: Added tests for combinations, some produce warning of objective function increasing but it's irrelevant because it's random data. 

PS: yes `Int` centers fail as per previous comments from @nalimilan there was a revert to force the centers to be of type `AbstractFloat` as otherwise it required more type gymnastic that was likely desirable. Float32 centers with Float64 data work.

src/kmeans.jl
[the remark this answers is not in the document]
tlienart: Fine, I've added more parametrization to the type, I think it's irrelevant and makes the code clunkier but it doesn't bother me that much. See ed49946

tlienart:
Thanks for the reviews @alyst , I think (hope) I've applied all your suggestions

src/kmeans.jl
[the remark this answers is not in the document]
nalimilan: It's really an essential rule in Julia that for performance all fields should be concretely typed.

src/kmeans.jl
```diff
@@ -2,13 +2,17 @@
 
 #### Interface
 
-mutable struct KmeansResult{T<:Real} <: ClusteringResult
-    centers::Matrix{T}         # cluster centers (d x k)
+# T is the eltype(centers)
```
nalimilan: `T` no longer exists?

tlienart: Well we move around a lot. Before @alyst's comment it was `centers::AbstractMatrix{T} where T<:AbstractFloat` and now I guess the point was to move the `AbstractMatrix` out of the struct and into the parametrisation whence the `C<:AbstractMatrix{<:AbstractFloat}`.

I think at this point I'm a bit unclear what to do if different than that so would appreciate hints to specific alternatives if one such is desired :)

tlienart: ha or maybe you're just commenting about the comment itself... I'll fix that

tlienart:
should we try to push this over the finish line? it's been long but it feels like most concerns have been addressed?

alyst:
@tlienart Yes, I think it should be go to go. I will just take one final until the end of the week.

src/kmeans.jl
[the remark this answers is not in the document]
alyst: Hmm, this line is still here, could you please remove it?

src/seeding.jl
```diff
@@ -40,22 +40,21 @@ initseeds_by_costs(algname::Symbol, costs::AbstractMatrix{<:Real}, k::Integer) =
...
-    d = size(X, 1)
-    n = size(X, 2)
+function copyseeds!(S::Matrix{<:AbstractFloat}, X::AbstractMatrix{<:Real},
+                    iseeds::AbstractVector)
+    d, n = size(X)
     k = length(iseeds)
-    (size(X,1) == d && size(S) == (d, k)) ||
-        throw(DimensionMismatch("Inconsistent array dimensions."))
-
+    size(S) == (d, k) || throw(DimensionMismatch("Inconsistent array " *
```
alyst: There's two spaces in the message after your change. I think it's fine to keep it on the single line (~80 chars)

```suggestion
    size(S) == (d, k) || throw(DimensionMismatch("Inconsistent array dimensions."))
```

alyst:
@tlienart Could you please add the two tiny changes I've noted? Besides that, I think it's ready to be merged.
@nalimilan There's #147, which is conflicting with this PR, but this one was created earlier and came through so many iterations that I think it would be fair if it is merged first.

tlienart:
Done @alyst , thanks!

nalimilan:
> @nalimilan There's #147, which is conflicting with this PR, but this one was created earlier and came through so many iterations that I think it would be fair if it is merged first.

Fine, anyhow rebasing shouldn't be hard. But CI currently fails for this PR.

tlienart:
The failure is unrelated to the PR, I think it’s due to changes in Distances, #147 also fails on the same I think.

What do you want me to do? I can try merging the changes of 147 here along with changes to fix the failure related to distances but that will protract the merge..

nalimilan:
OK, since what you did is essentially a rebased, I've merged #147 and finished the rebasing. It would have been weird to merge this PR first with changes from #147 which are completely unrelated to this PR.

tlienart:
So I guess this can be merged now? thanks both for all the reviews!

alyst:
@tlienart Sorry for the delay; finally merged! :) Thanks for your hard work! Now that we've converged on the requirements for the clustering methods refactoring, you are welcome to "refresh" the other ones. I think it brings the code base up to the current Julia standards and improves the maintainability in the long run.

PR #147, opened 2019-02-27
Add 'dims=2' to all 'pairwise' and 'pairwise!' calls

### LINKED ISSUE #146: Warnings with v12.2:  pairwise(…) should be replaced by pairwise(…, dims=2)
Just for reference (it's only a warning yet, but I guess it will become an error at some point):
```
┌ Warning: implicit `dims=2` argument now has to be passed explicitly to specify that distances between columns should be computed
│   caller = _kmeans!(::Array{Float64,2}, ::Nothing, ::Array{Float64,2}, ::Array{Int64,1}, ::Array{Float64,1}, ::Array{Int64,1}, ::Array{Float64,1}, ::Int64, ::Float64, ::Int64, ::Distances.SqEuclidean) at kmeans.jl:115
└ @ Clustering ~/.julia/packages/Clustering/pvAp6/src/kmeans.jl:115

```

Using Clustering.jl v12.2, Distances 0.8.0,  julia 1.1.0.

It looks like it comes from the calls to `pairwise` and `pairwise!` from the Distances.jl package. I think I got only the first warning, because it looks like there are a few different places where these functions are called.

Djoop:
Fixes #146 . Some tests fail, but it is already the case on master.

### DIFF
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -82,7 +82,7 @@ function _kmeans!(
     unused = Vector{Int}()
     num_affected::Int = k # number of centers, to which the distances need to be recomputed
 
-    dmat = pairwise(distance, centers, X)
+    dmat = pairwise(distance, centers, X; dims=2)
     dmat = convert(Array{T}, dmat) # Can be removed if one day Distance.result_type(SqEuclidean(), T, T) == T
     update_assignments!(dmat, true, assignments, costs, counts, to_update, unused)
     objv = w === nothing ? sum(costs) : dot(w, costs)
@@ -112,11 +112,11 @@ function _kmeans!(
         end
 
         if t == 1 || num_affected > 0.75 * k
-            pairwise!(dmat, distance, centers, X)
+            pairwise!(dmat, distance, centers, X; dims=2)
         else
             # if only a small subset is affected, only compute for that subset
             affected_inds = findall(to_update)
-            dmat_p = pairwise(distance, centers[:, affected_inds], X)
+            dmat_p = pairwise(distance, centers[:, affected_inds], X, dims=2)
             dmat[affected_inds, :] .= dmat_p
         end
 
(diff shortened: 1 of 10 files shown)


test/kmeans.jl
```diff
@@ -11,7 +11,8 @@ mutable struct MySqEuclidean <: SemiMetric end
 
 # redefinition of Distances.pairwise! for MySqEuclidean type
 function pairwise!(r::AbstractMatrix, dist::MySqEuclidean,
-                   a::AbstractMatrix, b::AbstractMatrix)
+                   a::AbstractMatrix, b::AbstractMatrix; dims::Integer)
+    dims == 2 || throw(ArgumentError("only dims=2 supported for MySqEuclidean distance"))
```
nalimilan: How about supporting this while you're at it? It shouldn't be hard (switch `tranpose` and `dims`, add tests after transposing).

src/kmeans.jl
```diff
@@ -82,7 +82,7 @@ function _kmeans!(
     unused = Vector{Int}()
     num_affected::Int = k # number of centers, to which the distances need to be recomputed
 
-    dmat = pairwise(distance, centers, X)
+    dmat = pairwise(distance, centers, X; dims=2)
```
```suggestion
    dmat = pairwise(distance, centers, X, dims=2)
```

src/kmeans.jl
```diff
@@ -112,11 +112,11 @@ function _kmeans!(
         end
 
         if t == 1 || num_affected > 0.75 * k
-            pairwise!(dmat, distance, centers, X)
+            pairwise!(dmat, distance, centers, X; dims=2)
```
```suggestion
            pairwise!(dmat, distance, centers, X, dims=2)
```

test/kmeans.jl
Djoop: Yes I agree it's easy to change, but what it is useful for? As far as I understand, this metric is defined only for testing, and the case dims=1 would never be called, am I wrong?

nalimilan: Sorry I hadn't noticed this was in tests.

YoungFaithful:
Looks great. Merging that would be especially helpful as the warnings slow the kmeans-algorithm down a lot.

PR #156, opened 2019-03-28
Specialize MCL functions for Sparse Matrices

altre:
I've needed to add some specialized functions for Sparse Matrices in the mcl implementation  to make it feasible for use on larger Matrices.
Benchmarking the old vs the new using:
```
using SparseArrays
using Clustering
using BenchmarkTools
using Random
Random.seed!(1)
adj = sprand(1000, 1000, 0.1)
@benchmark mcl(adj)
```
Old:
```julia> @benchmark mcl(adj)
BenchmarkTools.Trial: 
  memory estimate:  251.94 MiB
  allocs estimate:  103199
  --------------
  minimum time:     54.339 s (0.35% GC)
  median time:      54.339 s (0.35% GC)
  mean time:        54.339 s (0.35% GC)
  maximum time:     54.339 s (0.35% GC)
  --------------
  samples:          1
  evals/sample:     1
```
New:
```
julia> @benchmark mcl(adj)
BenchmarkTools.Trial: 
  memory estimate:  48.41 MiB
  allocs estimate:  20675
  --------------
  minimum time:     144.370 ms (5.11% GC)
  median time:      228.582 ms (9.42% GC)
  mean time:        279.637 ms (16.85% GC)
  maximum time:     578.249 ms (26.47% GC)
  --------------
  samples:          18
  evals/sample:     1
```

Note that this is a contrived example, using a dense matrix of size (1000,1000) is faster. Using this only makes sense for larger sparse matrices which don't fit into memory.

### DIFF
--- a/src/mcl.jl
+++ b/src/mcl.jl
@@ -5,7 +5,7 @@
 
 Result returned by `mcl()`.
 """
-struct MCLResult 
+struct MCLResult <: ClusteringResult
     mcl_adj::AbstractMatrix     # final MCL adjacency matrix (equilibrium state matrix if converged)
     assignments::Vector{Int}    # element-to-cluster assignments (n)
     counts::Vector{Int}         # number of samples assigned to each cluster (k)
--- a/src/mcl.jl
+++ b/src/mcl.jl
@@ -97,7 +97,7 @@ function _mcl_inflate(src::AbstractMatrix, inflation::Number)
     dest
 end
 
-function _mcl_inflate(dest::SparseMatrixCSC, src::SparseMatrixCSC, inflation::Number)
+function _mcl_inflate(src::SparseMatrixCSC, inflation::Number)
     dest = similar(src)
     @inbounds for i in 1:length(src.nzval)
         dest.nzval[i] = _mcl_el_inflate(src.nzval[i], inflation)
--- a/src/mcl.jl
+++ b/src/mcl.jl
@@ -5,7 +5,7 @@
 
 Result returned by `mcl()`.
 """
-struct MCLResult <: ClusteringResult
+struct MCLResult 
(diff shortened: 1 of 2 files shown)


alyst:
If it's possible to speed up MCL in the sparse case, that would be nice! I was worried that expansion will generate too many nonzeros for any sparse matrix and outweigh any performance gains, but I might be wrong. Could you please add tests checking that a) sparse case works (doesn't crash), b) the results of sparse and dense implementations are identical (for some reasonable nontrivial case)? We will need that to accept the PR, but it would also help you to shape the changes in the MCL code.

altre:
The concrete case we are working on is a 70000 x 70000 matrix with 0.3 % sparsity, which takes a while, but finishes in reasonable time.

There is already a 'doesn't crash' test in the test suite for sparse matrices. Would you say it's enough to just use the input to this test and compare the dense and sparse results?

alyst:
> There is already a 'doesn't crash' test in the test suite for sparse matrices. Would you say it's enough to just use the input to this test and compare the dense and sparse results?

That one is really a "doesn't crash" test. We need something more elaborate.
One possibility is to generate random datapoints from a mixture of K normal distributions (i.e. what Clustering.jl logo shows) -- the benefit is that it gives us the expected labels for the datapoints.
But it doesn't produce sparse adjacency matrices, so better ideas are welcome.

Also, it would be nice to add MCL (both dense and sparse) into the `benchmarks/benchmarks.jl`.
We would need the benchmarks for this PR anyway (for the dense branch too, since your changes affect it, e.g. `mcl_inflate()` is not inplace anymore), and extending `benchmarks.jl` looks like the best strategy.

There was some refactoring of the docs, updates to the `mcl()` params on the master + I fixed broken `mcl()` logging, so you would have to rebase the branch. Sorry for the inconvenience.

altre:
I will close this for now, and reopen if I get around to it

PR #166, opened 2019-04-12
Cleanup clustering evaluation

alyst:
The PR introduces `crosstab(clustering1, clustering2)` method and uses it in `varinfo()`, `randindex()` and `vmeasure()` to reduce code duplication and enhance parameter checks;
plus `varinfo()` calculation is simplified.

One question for RFC is the name of the new function (`crosstab`).
It's just a wrapper over `StatsBase.counts(a, b, 1:maximum(a), 1:maximum(b))`, i.e. the ranges of counted elements are always hardcoded to be from 1 to the maximal cluster number.
In principle, it would also be possible just to extend `StatsBase.counts` for `::ClusteringResult` arguments without introducing a new method, but then the users would always have to specify the elements range (`1:maximum(a)`) when calling it for assignment vectors.
The alternative name suggestions are also welcome. (E.g. `contingencymatrix()`).


nalimilan:
I agree `crosstab` sounds too general for Clustering.jl to claim this name. `crosstab` would really be the same thing as `StatsBase.counts` or `FreqTables.freqtable, but with a different name just to avoid conflicts or dependencies. I think the `StatsBase.counts` API needs to be improved (it only accepts a single array currently, and its name doesn't indicate it only works for integer values), but it sounds simpler to overload that unless you want to depend on FreqTables.

> In principle, it would also be possible just to extend `StatsBase.counts` for `::ClusteringResult` arguments without introducing a new method, but then the users would always have to specify the elements range (`1:maximum(a)`) when calling it for assignment vectors.

Why couldn't we use the appropriate default? We can do whatever we want in the `::ClusteringResult` method.

alyst:
I also prefer extending `StatsBase.counts()` with
```
StatsBase.counts(a::ClusteringResult, b::ClusteringResult) = counts(assignments(a), assignments(b), 1:maximum(a), 1:maximum(b))
StatsBase.counts(a::ClusteringResult, b::IntegerArray) = counts(assignments(a), b, 1:maximum(a), 1:maximum(b))
StatsBase.counts(a::IntegerArray, b::ClusteringResult) = counts(a, assignments(b), 1:maximum(a), 1:maximum(b))
```
which is fine.

The problem is with
```
StatsBase.counts(a::IntegerArray, b::IntegerArray) = counts(a, b, (span(a), span(b)))
```
as it is defined in StatsBase.
`span(a) == extrema(a)`, so if the user specifies assignments array with cluster indices not starting from 1, indexing the result of `counts(a, b)` by the `a` cluster numbers would give incorrect results.
IIUC that should not affect the clustering measures that depend on `counts()`. But the direct use of `counts()` might be confusing.

We can as well just use `span(a/b)` everywhere -- at the expense of breaking indexing by cluster numbers in some corner cases (user-provided assignments arrays).

nalimilan:
Maybe only define the first three methods, and leave the last one as-is?

alyst:
As discussed, `crosstab()` was changed into `StatsBase.counts()`.
In addition to `1:maximum(a)` vs `span(a)` indexing difference between `ClusteringResult`-specific and generic `counts()` methods, there's another one: the generic `counts()` throws `ArgumentError` when empty array is given, while the Clustering version would return 0x0 matrix.

I have also merged the evaluation&validation doc pages into a single one: they were quite short, so the change will make the navigation faster.

Issue #107, opened 2017-12-13
Implement fast optimal leaf ordering for Hclust


kescobo:
I recently had need of an implementation of the method in [this paper](https://www.ncbi.nlm.nih.gov/pubmed/11472989):

>We present the first practical algorithm for the optimal linear leaf ordering of trees that are generated by hierarchical clustering. Hierarchical clustering has been extensively used to analyze gene expression data, and we show how optimal leaf ordering can reveal biological structure that is not observed with an existing heuristic ordering method. For a tree with n leaves, there are 2(n-1) linear orderings consistent with the structure of the tree. Our optimal leaf ordering algorithm runs in time O(n(4)), and we present further improvements that make the running time of our algorithm practical.

I'm not sure I did it the most efficient way possible, but for an hclust of a 5k x 5k distance matrix it took ~50 ms (generating the hclust itself was ~1000 ms). [See jupyter notebook here](https://gist.github.com/kescobo/48ef708c932f2cd9894cdebbea75b487) 

I initially wrote it for my Microbiome package, but I think it makes more sense to live here if you're up for a PR

PR #170, opened 2019-06-10
Add optimal ordering methods (#107)

kescobo:
Based on:
[Bar-Joseph et. al. "Fast optimal leaf ordering for hierarchical clustering." _Bioinformatics_. (2001)](https://doi.org/10.1093/bioinformatics/17.suppl_1.S22) (addresses #107)

To see it in action:

```
using Clustering
using Distances
using StatsPlots
using Random

n = 40

mat = zeros(Int, n, n)

for i in 1:n
   last = minimum([i+Int(floor(n/5)), n])
   for j in i:last
       mat[i,j] = 1
   end
end

mat = mat[:, randperm(n)]
heatmap(mat)
```

Screen Shot 2019-06-10 at 5 27 23 PM

```
dm = pairwise(Euclidean(), mat, dims=2)
hcl = hclust(dm, linkage=:average)
heatmap(mat[:,hcl.order])
```

Screen Shot 2019-06-10 at 5 27 46 PM

```
hcl2 = optimalorder(hcl, dm)
heatmap(mat[:,hcl2.order])
```

Screen Shot 2019-06-10 at 5 28 39 PM


kescobo:
I've been playing around with some ways to test this, but not coming up with anything super useful (aside from making sure that `Hclust` objects are returned from both functions and similar trivial things). Suggestions welcome.

alyst:
@kescobo Thanks, I think it would be very useful! I am not familiar with the algorithm, does the leaves reordering imply that the merges structure has to be updated as well?
Does that mean that potentially it could be implemented as an alternative to `rorder!()` function (that reorders merges to match R output)?
Say, we introduce `leaforder::Symbol` keyword argument to `hclust()` function that could be either `:r` or `:barjoseph` (`:optimal` as an alias) and then call either `rorder!()` or `barjosephorder!()` when constructing `Hclust`.
We would not allow reordering after `Hclust` is created, making sure that `.merges` and `.order` are in sync.
That way we also avoid the problem that `dm` may not match the original distance matrix.

kescobo:
>does the leaves reordering imply that the merges structure has to be updated as well?

I don't _think_ so. The reason is that the order of merges is based on the distance of elements. So if `a` and `b` are the closest, they will be the first merge. This algorithm doesn't alter anything about the  structure of the tree except which leaves are next to each other (it's rotating nodes, not changing where the nodes are).

To double check, I plotted a smaller number so they can be inspected manually:

```
julia> n = 20; mat = zeros(Int, n, n);

julia> for i in 1:n
          last = minimum([i+Int(floor(n/5)), n])
          for j in i:last
              mat[i,j] = 1
          end
       end

julia> dm = pairwise(Euclidean(), mat, dims=2);

julia> hcl = hclust(dm, linkage=:average);

julia> plot(hcl, xticks=(1:20, "abcdefghijklmnopqrst"[hcl.order]))

julia> heatmap(mat[:, hcl.order], colorbar=false, xticks=(1:20, "abcdefghijklmnopqrst"[hcl.order]))
```

unordered

unordered_heat


```
julia> hcl2 = optimalorder(hcl, dm);

julia> plot(hcl2, xticks=(1:20, "abcdefghijklmnopqrst"[hcl2.order]))

julia> heatmap(mat[:, hcl2.order], colorbar=false, xticks=(1:20, "abcdefghijklmnopqrst"[hcl2.order]))
```
ordered

ordered_heat

Then again, you're correct that in R, some of the merges are altered. So for example a `(-1, -2)` might be changed to `(-2, -1)`. But this is an arbitrary ordering right?

kescobo:
>Say, we introduce leaforder::Symbol keyword argument to hclust() function that could be either :r or :barjoseph (:optimal as an alias) and then call either rorder!() or barjosephorder!() when constructing Hclust.
>We would not allow reordering after Hclust is created, making sure that .merges and .order are in sync.
>That way we also avoid the problem that dm may not match the original distance matrix.

This seems sensible to me, though I'm not sure I understand the internals of hclust enough to implement myself. Can I just not export `optimalorder()` and then call it inside `hclust()` at the end with an `if leaforder == :optimal`?

alyst:
> This algorithm doesn't alter anything about the structure of the tree except which leaves are next to each other (it's rotating nodes, not changing where the nodes are).

Exactly, the sequence of merges is not altered, but left and right branches of the node may be flipped.
That's also what `rorder!()` does _(basically, it tries to make the order of elements matching their initial order (row and col index in the distance matrix), and make the order of branches matching their position in the `.merges`)_.
But that means that `.merges` have to be updated as well (e.g. `c` and `d` parent and `e` leaf are flipped after `optimalorder()`, and so must be corresponding `merges` entries).

> Can I just not export optimalorder() and then call it inside hclust() at the end with an if leaforder == :optimal?

`rorder!()` (I'd rename it into `orderleaves_r!()`) works with `HclustMerges` structure. It would be nice if `orderleaves_barjoseph!()` (I'd avoid calling it optimal in the sources, because there could be different definitions of optimality) works with it as well.
Then we can just call either `orderleaves_r!()` or `orderleaves_barjoseph!()` before the call to `Hclust` ctor in `hclust()`.

src/hclust.jl
```diff
@@ -787,3 +787,96 @@ function printupper(d::Matrix)
...
+        else
+            error("invalid 'merge' order in Hclust: ($vl, $vr) ")
+        end
+        push!(extents, (uidx, widx))
+    end
+end
+
+
+function leaflocs(v::Int, order::Vector{Int}, extents::Vector{Tuple{Int,Int}})
+    ##################
```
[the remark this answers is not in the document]
kescobo: Hmm - don't remember where that came from.

src/hclust.jl
```diff
@@ -787,3 +787,96 @@ function printupper(d::Matrix)
...
+    if v < 0
+        leftextent = findfirst(abs(v) .== order)
```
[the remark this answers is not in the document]
kescobo: Cool! OK to use `isequal(-v)` instead of `==(-v)`?

src/hclust.jl
```diff
@@ -787,3 +787,96 @@ function printupper(d::Matrix)
...
+        rightextent = leftextent
+    elseif v > 0
+        leftextent = extents[v][1]
+        rightextent = extents[v][2]
+    else
+        error("leaf position cannot be zero")
+    end
+    left = order[leftextent]
```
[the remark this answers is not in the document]
kescobo: true

kescobo:
Looking into the last bit now (fixing the merges). Am I understanding correctly that if the merges are correct, that `Hclust(merges, linkage)` will determine the order automatically?

alyst:
> Am I understanding correctly that if the merges are correct, that `Hclust(merges, linkage)` will determine the order automatically?

Yes, that's what `hclust_perm()` should be doing when you provide the proper `HclustMerges` object.

kescobo:
@alyst Sorry, I'm having trouble wrapping my head around the way merges work. I remember this being a problem when I first wrote this code a year or so ago, I had to draw everything out, but I've done that here and I'm not groking it. Perhaps you can help - 

Take as an example 

```
using Clustering, Distances, StatsPlots, Random
Random.seed!(2)
n = 10

mat = zeros(Int, n, n)

for i in 1:n
   last = minimum([i+Int(floor(n/5)), n])
   for j in i:last
       mat[i,j] = 1
   end
end

mat = mat[:, randperm(n)]
heatmap(mat)

dm = pairwise(Euclidean(), mat, dims=2)
hcl = hclust(dm, linkage=:average)

plot(hcl, xticks=(collect(1:n), ["$i" for i in hcl.order]))
```
Screen Shot 2019-06-18 at 2 38 55 PM

with merges

```
(-1, -6)
(-9,  1)
(-3, -5)
(-8, -10)
(-2, -4)
(-7,  5)
( 2,  4)
( 3,  7)
( 6,  8)
```

When I the code I originally pushed:

```
hcl2 = optimalorder(hcl, dm)
plot(hcl2, xticks=(collect(1:n), ["$i" for i in hcl2.order]))
```

Should the 6th merge be `(5, -7)`? as far as I can tell, currently in any leaf:cluster pair in merges, the leaf (negative number) always comes first. In fact, I think I wrote the plotting recipe in a way that assume this. Is that just the way R handles things, or something more fundamental?

kescobo:
I also don't totally understand how the `HclustMerge()` is created / how I should alter my current code to modify it. Any pointers would be appreciated.

kescobo:
As a first step, I'm just trying to see if I can get the merges reordered at the end rather than working upstream (I'd like to build some tests on something that works before diving into the other data structures), but I can't figure out why this code isn't working.

alyst:
>Should the 6th merge be (5, -7)? as far as I can tell, currently in any leaf:cluster pair in merges, the leaf (negative number) always comes first. In fact, I think I wrote the plotting recipe in a way that assume this. Is that just the way R handles things, or something more fundamental?

To match the plot it should be `(-7, 5)`. `(leaf, cluster)` is the R order set by `rorder!()` (see also `_isrorder()` method), nothing "fundamental".

> I also don't totally understand how the HclustMerge() is created / how I should alter my current code to modify it. Any pointers would be appreciated.

`HclustMerges` is very similar to `Hclust`. The idea is that `Hclust` is exported public struct, it is created "at once" and is not meant to be modified. OTOH `HclustMerges` is an internal struct and has internal API to grow the tree (`push_merge()`). The fields are very similar, just instead of _n x 2_ `Hclust.merges` matrix (which, I guess, is there to match R Hclust class)  there are `mleft` and `mright` vectors (which IMHO is slightly easier to work with).

`HclustMerges` does not have 'order` field, but you can compute it within `orderleaves_barjoseph(hmer::HclustMerges)` as `order = invperm(hclust_perm(hmer))` and use it instead of `hcl.order`. The only other thing you have to change is to use `hmer.mleft` and `hmer.mright` instead of `hcl.merges[:, 1]` and `hcl.merges[:, 2]`.

src/hclust.jl
```diff
@@ -531,6 +531,101 @@ function rorder!(hmer::HclustMerges)
...
+        rightextent = extents[v][2]
+    else
+        error("leaf position cannot be zero")
+    end
+
+    return order[leftextent], order[rightextent], leftextent, rightextent
+end
+
+# recursively rotate merges
+function rotate_merges!(merges, idx)
```
alyst:
```suggestion
function rotate_merges!(hmer::HclustMerges, i::Integer)
```

src/hclust.jl
```diff
@@ -543,7 +638,7 @@ end
 ##   until c[i] = c[i-2] ## nearest of nearest is cluster itself
 ##   merge c[i] and nearest neighbor c[i]
 ##   if i>3 i -= 3 else i <- 1
-function hclust_nn(d::AbstractMatrix, linkage::Function)
+function hclust_nn(d::AbstractMatrix, linkage::Function, leaforder=:r)
```
alyst:
```suggestion
function hclust_nn(d::AbstractMatrix, linkage::Function)
```

src/hclust.jl
```diff
@@ -590,14 +685,14 @@ function hclust_nn(d::AbstractMatrix, linkage::Function)
         end
         isempty(NN) && push!(NN, 1) # restart NN chain
     end
-    return rorder!(htre.merges)
+    return orderleaves_r!(htre.merges)
```
alyst:
```suggestion
    return htre.merges
```

At the moment all specific hierarchical clustering methods (`hclust_xxx()`) call `rorder!()` at the end.
Since we are introducing alternative ordering options, we should stop doing this and call `orderleaves_r!()` once -- from `hclust()`.

src/hclust.jl
```diff
@@ -691,7 +793,11 @@ function hclust(d::AbstractMatrix; linkage::Symbol = :single,
         throw(ArgumentError("Unsupported cluster linkage $linkage"))
     end
 
-    Hclust(hmer, linkage)
+    hcl = Hclust(hmer, linkage)
+    if leaforder == :barjoseph
+        orderleaves_barjoseph!(hcl, d)
+    end
```
alyst:
```suggestion
    elseif leaforder == :r
        orderleaves_r!(hmer) 
    end
```

src/hclust.jl
```diff
@@ -691,7 +793,11 @@ function hclust(d::AbstractMatrix; linkage::Symbol = :single,
         throw(ArgumentError("Unsupported cluster linkage $linkage"))
     end
 
-    Hclust(hmer, linkage)
+    hcl = Hclust(hmer, linkage)
```
alyst: This should be done after applying leaves order.

src/hclust.jl
```diff
@@ -531,6 +531,101 @@ function rorder!(hmer::HclustMerges)
...
+
+Sub branches are flipped if necessary to minimize the distance between adjacent
+nodes, and then the combined branches are treated as a block for future
+comparisons.
+
+Based on:
+[Bar-Joseph et. al. "Fast optimal leaf ordering for hierarchical clustering." _Bioinformatics_. (2001)](https://doi.org/10.1093/bioinformatics/17.suppl_1.S22)
+"""
+function orderleaves_barjoseph!(hcl::Hclust, dm::Array{Float64,2})
+    extents = Tuple{Int,Int}[]
```
alyst: I'd rather call it 
```suggestion
    node_ranges = Tuple{Int,Int}[] # ranges of order array indices occupies by the leaves of the given node 
```

src/hclust.jl
```diff
@@ -531,6 +531,101 @@ function rorder!(hmer::HclustMerges)
...
+compared (a maximum of 4 comparisons per intersection).
+
+Sub branches are flipped if necessary to minimize the distance between adjacent
+nodes, and then the combined branches are treated as a block for future
+comparisons.
+
+Based on:
+[Bar-Joseph et. al. "Fast optimal leaf ordering for hierarchical clustering." _Bioinformatics_. (2001)](https://doi.org/10.1093/bioinformatics/17.suppl_1.S22)
+"""
+function orderleaves_barjoseph!(hcl::Hclust, dm::Array{Float64,2})
```
alyst:
```suggestion
function orderleaves_barjoseph!(hmer::HclustMerges, d::AbstractMatrix)
```
Note that `d` might be different from `Matrix{Float64}`.

src/hclust.jl
```diff
@@ -531,6 +531,101 @@ function rorder!(hmer::HclustMerges)
...
+    if v < 0
+        leftextent = rightextent = findfirst(isequal(-v), order)
+    elseif v > 0
+        leftextent = extents[v][1]
+        rightextent = extents[v][2]
+    else
+        error("leaf position cannot be zero")
+    end
+
+    return order[leftextent], order[rightextent], leftextent, rightextent
```
alyst: Let's simplify it and just return `left, right`. We can do `order[left]`, `order[right]` in the `orderleaves_barjoseph!()` directly.
IMHO it will make the code easier to understand.

src/hclust.jl
```diff
@@ -531,6 +531,101 @@ function rorder!(hmer::HclustMerges)
...
+    leaflocs(v::Int, order::Vector{Int}, extents::Vector{Tuple{Int,Int}})
+
+`v`: vertex - may be a leaf (negative numbers) or a merge index
+`order`: Vector of leaf positions, same as hclust.order
+`extents`: tuples of the order indices for outermost leaves for each merge
+
+Returns the `order` values and indices of the extents for a given `v`.
+If `v` is a leaf, left and right extents will be the same.
+"""
+function leaflocs(v::Int, order::Vector{Int}, extents::Vector{Tuple{Int,Int}})
```
alyst: Maybe rename it into something like (along with simplifying the returned values):
```suggestion
function leaves_range(v::Int, order::Vector{Int}, node_ranges::Vector{Tuple{Int,Int}})
```

kescobo:
Something I changed altered something unintended, and now I'm consistently getting one thing failing to flip

Screen Shot 2019-06-20 at 10 55 39 AM

kescobo:
@alyst Thanks for another extensive review! I think the problem was that I revered the left and right in `rotate_merges!()`. 

Now, however, a bunch of tests are failing, and they all seem to be related to `hclust_nn()`. I thought this might be because `rorder!()` is *not* called when the `linkage=:single`, since `hclust_minumum()` doesn't call it and now we call it at the end of `hclust()`, but I added

```
    elseif linkage == :single
        # do nothing
```

[here](https://github.com/JuliaStats/Clustering.jl/pull/170/files#diff-7ad84cca36ddae742d57f7adc00f4d91R805) and the tests still fail. Any thoughts?

src/Clustering.jl
```diff
@@ -58,7 +58,7 @@ module Clustering
     vmeasure,
 
     # hclust
-    Hclust, hclust, cutree,
+    Hclust, hclust, cutree, optimalorder, optimalorder!,
```
alyst:
```suggestion
    Hclust, hclust, cutree,
```

src/hclust.jl
```diff
@@ -509,7 +509,7 @@ end
 
 ## reorders the tree merges by the height of resulting trees
 ## (to be compatible with R's hclust())
-function rorder!(hmer::HclustMerges)
+function orderleaves_r!(hmer::HclustMerges)
```
alyst: Before we merge this: should we rather call these methods `orderbranches_xxx!`?
This sounds more correct to me. Sorry for extra work.

src/hclust.jl
```diff
@@ -531,6 +531,108 @@ function rorder!(hmer::HclustMerges)
...
+                reverse!(order, kidx, widx)
+                rotate_merges!(hmer, vr)
+            end
+        else
+            error("invalid 'merge' order in Hclust: ($vl, $vr) ")
+        end
+
+        push!(node_ranges, (uidx, widx))
+    end
+    @show order
```
alyst:
```suggestion
    return hmer
```

src/hclust.jl
```diff
@@ -531,6 +531,108 @@ function rorder!(hmer::HclustMerges)
...
+            if flp == 2 || flp == 4
+                reverse!(order, uidx, midx)
+                rotate_merges!(hmer, vl)
+            end
+            if flp == 3 || flp == 4
+                reverse!(order, kidx, widx)
+                rotate_merges!(hmer, vr)
+            end
+        else
+            error("invalid 'merge' order in Hclust: ($vl, $vr) ")
```
alyst:
```suggestion
            error("invalid 'merge' order in Hclust: ($vl, $vr)")
```
Would this be triggered only when `vl == vr == 0`?

src/hclust.jl
```diff
@@ -531,6 +531,108 @@ function rorder!(hmer::HclustMerges)
...
+end
+
+"""
+    node_range(v::Int, order::Vector{Int}, node_ranges::Vector{Tuple{Int,Int}})
+
+`v`: vertex - may be a leaf (negative numbers) or a merge index
+`order`: Vector of leaf positions, same as hclust.order
+`extents`: tuples of the order indices for outermost leaves for each merge
+
+Returns the `order` values and indices of the extents for a given `v`.
```
alyst:
```suggestion
Returns the left and right bounds of the range.
```

src/hclust.jl
```diff
@@ -531,6 +531,108 @@ function rorder!(hmer::HclustMerges)
...
+    end
+    @show order
+end
+
+"""
+    node_range(v::Int, order::Vector{Int}, node_ranges::Vector{Tuple{Int,Int}})
+
+`v`: vertex - may be a leaf (negative numbers) or a merge index
+`order`: Vector of leaf positions, same as hclust.order
+`extents`: tuples of the order indices for outermost leaves for each merge
```
alyst:
```suggestion
`node_ranges`: left and right bounds of the `order` array indices occupied by each node
```

src/hclust.jl
kescobo: While it *does* order the branches, in my mind the important part is the order of the leaves. Ordering of the branches is incidental. But also I don't really feel strongly about it. Would you want to change the arg in `hclust()` to `branchorder=:r` ?

src/hclust.jl
```diff
@@ -531,6 +531,108 @@ function rorder!(hmer::HclustMerges)
...
+        end
+
+        push!(node_ranges, (uidx, widx))
+    end
+    @show order
+end
+
+"""
+    node_range(v::Int, order::Vector{Int}, node_ranges::Vector{Tuple{Int,Int}})
+
```
alyst:
```suggestion
Get the left and right bounds of the range of the `order` array indices occupied by the elements of the `v` node.
```

Actually, we don't need to be more verbose than that, since it's an internal function (we even don't need to have it as a docstring, the comment would have been also fine).

src/hclust.jl
alyst: > Would you want to change the arg in hclust() to branchorder=:r ?

Yes, it should be in sync.

src/hclust.jl
```diff
@@ -662,9 +764,16 @@ Returns the dendrogram as a [`Hclust`](@ref) object.
...
+   The valid choices are:
+   * `:r` (the default): ordered based on the height of merges (compatible) with
+     R's `hclust`
+   * `:barjoseph`: leaves are ordered to reduce the distance between neighboring
+     leaves from separate branches using the "fast optimal leaf ordering" algorithm
+     from Bar-Joseph et. al. _Bioinformatics_ (2001)
 """
 function hclust(d::AbstractMatrix; linkage::Symbol = :single,
-                uplo::Union{Symbol, Nothing} = nothing)
+                uplo::Union{Symbol, Nothing} = nothing, leaforder=:r)
```
alyst:
```suggestion
                uplo::Union{Symbol, Nothing} = nothing, leaforder::Symbol=:r)
```

src/hclust.jl
kescobo: Would you prefer a comment? I will definitely not remember how these things work in 3 months without being more verbose

src/hclust.jl
```diff
@@ -662,9 +764,16 @@ Returns the dendrogram as a [`Hclust`](@ref) object.
  - `uplo::Symbol` (optional): specifies whether the upper (`:U`) or the
    lower (`:L`) triangle of `d` should be used to get the distances.
    If not specified, the method expects `d` to be symmetric.
+- `leaforder::Symbol` (optional): algorithm to determine ordering of leaves.
+   The valid choices are:
+   * `:r` (the default): ordered based on the height of merges (compatible) with
```
alyst:
```suggestion
   * `:r` (the default): ordered based on the node heights and original elements order (compatible with R's `hclust`)
```

src/hclust.jl
alyst: > Would you prefer a comment?

Yes, the reason is that maintaining docstring (e.g. syncing arg names and types) is an extra work.

>  I will definitely not remember how these things work in 3 months without being more verbose

Same is true for me :) But then I'd rather add the comments in the method body, since it's an internal function. E.g. `v < 0 # the leaf node`

alyst:
@kescobo Thanks for addressing the comments!

> Now, however, a bunch of tests are failing, and they all seem to be related to hclust_nn(). 

Yes, that's because the tests are calling an alternative NN implementation (`hclust_nn`) directly (to check that the faster one from `hclust()` and `hclust_nn()` give the same results).
Just change the L54 in tests/hclust.jl into
```julia
linkage_fun = linkage == :complete ? Clustering.slicemaximum : Clustering.slicemean
hclu2 = Hclust(Clustering.orderleaves_r!(Clustering.hclust_nn(example["D"], linkage_fun)), linkage)
```
That should work.

src/hclust.jl
```diff
@@ -691,6 +801,13 @@ function hclust(d::AbstractMatrix; linkage::Symbol = :single,
         throw(ArgumentError("Unsupported cluster linkage $linkage"))
     end
 
+    if leaforder == :barjoseph
```
alyst: Let's also support `:optimal` as an alias (please update the dockstring):
```suggestion
    if leaforder == :barjoseph || leaforder == :optimal
```

It could be that we'll find the "more optimal" algorithm, then we'll be update to change what `:optimal` does while keeping `:barjoseph`.

test/hclust.jl
```diff
@@ -141,4 +142,27 @@ end
...
+        for j in i:last
+            mat[i,j] = 1
+        end
+    end
+
+    dm = pairwise(Euclidean(), mat, dims=2)
+
+    hcl1 = hclust(dm, linkage=:average)
+    hcl2 = hclust(dm, linkage=:average, leaforder=:barjoseph)
+
```
alyst: Please also add the test for `:optimal` and for some incorrect value
```julia
@test_throws ArgumentError hclust(dm, linkage=:average, leaforder=:wrong)
```
And the tests with 0x0 and 1x1 distance matrices.

test/hclust.jl
```diff
@@ -141,4 +142,27 @@ end
...
+    for i in 1:n
+        last = minimum([i+Int(floor(n/5)), n])
+        for j in i:last
+            mat[i,j] = 1
+        end
+    end
+
+    dm = pairwise(Euclidean(), mat, dims=2)
+
+    hcl1 = hclust(dm, linkage=:average)
```
alyst: Maybe `hcl_r` and `hcl_bj` for clarity?

kescobo:
🎉 OK, I think that's everything, and tests are passing locally.

But since I was editing as you were commenting, I almost missed a couple of things. I scanned back through and think I got through them all, but please let me know if I missed something. 

Regardless, thanks for taking so much time to review and improve this PR!

src/hclust.jl
```diff
@@ -662,9 +755,16 @@ Returns the dendrogram as a [`Hclust`](@ref) object.
  - `uplo::Symbol` (optional): specifies whether the upper (`:U`) or the
    lower (`:L`) triangle of `d` should be used to get the distances.
    If not specified, the method expects `d` to be symmetric.
+- `branchorder::Symbol` (optional): algorithm to determine ordering of leaves.
+   The valid choices are:
+   * `:r` (the default): ordered based on the node heights and original elements
+     order (compatible with R's `hclust`)
+   * `:barjoseph` (or :optimal): leaves are ordered to reduce the distance between neighboring
```
alyst:
```suggestion
   * `:barjoseph` (or `:optimal`): branches are ordered to reduce the distance between neighboring
```

src/hclust.jl
kescobo: Yes, I think that's right. I think when I was originally writing this code, I didn't cover the case where `vl` was positive and `vr` was negative because it seemed like that never happened in merges. I added this `else` to make sure that never happened. Would you prefer I remove it?

The only harm I think is that it hurts codecov percentage :-/

src/hclust.jl
```diff
@@ -531,6 +531,99 @@ function rorder!(hmer::HclustMerges)
...
+        if vl < 0 && vr < 0
+            # Nothing needs to be done
+        elseif vl < 0
+            # check if flipping would reduce distance
+            if dm[m,k] > dm[m,w]
+                reverse!(order, uidx, midx)
+                rotate_merges!(hmer, vr)
+            end
+        elseif vr < 0
+            if dm[k,m] > dm[k,u]
```
alyst: coveralls reports that this line is not covered by the tests. Do you have some example in mind that would test this line? (Maybe similar to the current band-matrix example, just with different number of elements?)
I still didn't dig into the algorithm details, but I wonder, why both in the `vl < 0` and `vr < 0` branches the same `uidx, midx` region is reversed.

alyst:
>  OK, I think that's everything, and tests are passing locally.

Yes, I think we're almost there. There's just that tiny docstring fix + maybe it's possible to devise the test to cover the line I commented about (that should fix the drop in the coverage).

> Regardless, thanks for taking so much time to review and improve this PR!

Thanks for patiently following the review! :) Your future contributions are very welcome!

src/hclust.jl
kescobo: >Do you have some example in mind that would test this line?

I could artificially construct some merges that do this, but I think this doesn't actually happen in the wild. I just tried making banded matrices for `n= 2:100`, as well as 1000 random matrices for each of those `n`s and that line is never triggered.

>I still didn't dig into the algorithm details, but I wonder, why both in the vl < 0 and vr < 0 branches the same uidx, midx region is reversed.

That's a mistake in copy-pasting. Should be `kidx` and `widx`

kescobo: See also [this comment](https://github.com/JuliaStats/Clustering.jl/pull/170#discussion_r295945252)

kescobo: Actually, I can't seem to artificially construct the `HclustMerges`... I think [these lines](https://github.com/JuliaStats/Clustering.jl/blob/master/src/hclust.jl#L99-L102) is what's preventing this from happening.

Should I just trust that this won't happen and remove the code? I can comment it out instead in case this ever changes.

src/hclust.jl
alyst: I think it's fine to keep this `error()`. I personally don't know a better alternative.
Maybe just add
```julia
else # vl == vr == 0
```
to help understanding the code.

src/hclust.jl
alyst: I think you're right. Could you pls try removing those lines from `push_merges!()`? Now that we support different orderings this flip will just cause confusion.

src/hclust.jl
```diff
@@ -557,6 +557,11 @@ function orderbranches_barjoseph!(hmer::HclustMerges, dm::AbstractMatrix)
         (u, m) = (order[uidx], order[midx])
         (k, w) = (order[kidx], order[widx])
 
+        # Make sure node order is correct
```
kescobo: What do you think of this as an alternative? This didn't cause any errors when I ran the tests locally, so it really seems like it can't happen.

src/hclust.jl
kescobo: Removing those lines caused 2 test failures:

```
  R hclust() generated examples                                             | 6549         6549
  hclust_n3()                                                               |    1     2      3
```

src/hclust.jl
alyst: Actually, I'd rather remove the swap from `push_merges!()` and also keep the error in the last `else` where it was before, because this check is rather artificial, and now it will happen for every node.

src/hclust.jl
alyst: That's no so bad :)
It's the same issue as with `hclust_nn()`.
Could you please wrap `hclust_n3()` call with `Clustering.orderbanches_r!()` in the tests at L84?

kescobo: Done!

src/hclust.jl
kescobo: Done! And confirmed that I now often hit `vr < 0` branch.

kescobo:
I'm really giving Travis a workout here.

src/hclust.jl
```diff
@@ -662,9 +752,16 @@ Returns the dendrogram as a [`Hclust`](@ref) object.
  - `uplo::Symbol` (optional): specifies whether the upper (`:U`) or the
    lower (`:L`) triangle of `d` should be used to get the distances.
    If not specified, the method expects `d` to be symmetric.
+- `branchorder::Symbol` (optional): algorithm to determine ordering of leaves.
```
alyst:
```suggestion
- `branchorder::Symbol` (optional): algorithm to order leaves and branches.
```

src/hclust.jl
```diff
@@ -662,9 +752,16 @@ Returns the dendrogram as a [`Hclust`](@ref) object.
...
+   The valid choices are:
+   * `:r` (the default): ordered based on the node heights and original elements
```
alyst:
```suggestion
   * `:r` (the default): ordering based on the node heights and the original elements
```

src/hclust.jl
```diff
@@ -531,6 +528,99 @@ function rorder!(hmer::HclustMerges)
...
+    return hmer
+end
+
+## Get the left and right bounds of the range of the `order` array indices occupied by the elements of the node.
+## If `node` is a leaf, left and right bounds will be the same.
+function node_range(node::Int, order::Vector{Int}, node_ranges::Vector{Tuple{Int,Int}})
+    if node < 0 # leaf node
+        left = right = findfirst(isequal(-node), order)
+    elseif node > 0 # branch node
+        left = node_ranges[node][1]
```
alyst:
```suggestion
        left, right = node_ranges[node]
```

src/hclust.jl
```diff
@@ -531,6 +528,99 @@ function rorder!(hmer::HclustMerges)
...
+        right = node_ranges[node][2]
```
alyst:
```suggestion
```

src/hclust.jl
```diff
@@ -662,9 +752,16 @@ Returns the dendrogram as a [`Hclust`](@ref) object.
  - `uplo::Symbol` (optional): specifies whether the upper (`:U`) or the
    lower (`:L`) triangle of `d` should be used to get the distances.
    If not specified, the method expects `d` to be symmetric.
+- `branchorder::Symbol` (optional): algorithm to determine ordering of leaves.
+   The valid choices are:
+   * `:r` (the default): ordered based on the node heights and original elements
+     order (compatible with R's `hclust`)
+   * `:barjoseph` (or :optimal): branches are ordered to reduce the distance between neighboring
```
alyst:
```suggestion
   * `:barjoseph` (or `:optimal`): branches are ordered to reduce the distance between neighboring
```

alyst:
:+1:  coveralls is green!
There's some tiny suggestions I've just added + maybe you can add the test for `hcl_rand.merges`.
Once it's done, I will merge the PR. I'm eager to test how it will look for the dataset I'm currently analyzing :)

kescobo:
Woo! Hopefully we can get this done today 🤞 

Just went through your most recent review, tests pass locally.

src/hclust.jl
```diff
@@ -662,9 +751,17 @@ Returns the dendrogram as a [`Hclust`](@ref) object.
  - `uplo::Symbol` (optional): specifies whether the upper (`:U`) or the
    lower (`:L`) triangle of `d` should be used to get the distances.
    If not specified, the method expects `d` to be symmetric.
+- `branchorder::Symbol` (optional): algorithm to determine order leaves and branches.
```
alyst: In my initial suggestion it was
```suggestion
- `branchorder::Symbol` (optional): algorithm to order leaves and branches.
```

kescobo: Ah sorry, github interface does this thing when I try to select that makes it hard to see. Copy-paste fail there

Screen Shot 2019-06-21 at 8 40 00 AM

alyst:
@kescobo All green! Thanks again for your dedication! :)

kescobo:
Hurray! It was my pleasure, thanks for yours :-).

PR #171, opened 2019-06-18
added the mutual information calculation

wildart:


src/mutualinfo.jl
```diff
@@ -0,0 +1,40 @@
...
+    hk = entAcols/N + log(N)
+
+    mi = hc - hck
+    nmi = 2*mi/(hc+hk)
+
+    return mi, nmi
+end
+
+"""
+    mutualinfo(a, b) -> Float64
```
alyst: Maybe we can add a keyword specifying whether we need the normalized MI?
```suggestion
    mutualinfo(a, b; normed::Bool = false) -> Float64
```

`multualinfo(a, b, normed=true)` is better than `mutualinfo(a, b)[2]`

src/mutualinfo.jl
```diff
@@ -0,0 +1,40 @@
+# Mutual Information
+
+function _mutualinfo(A::AbstractMatrix{<:Integer})
+    N = sum(A)
+    (N == 0.0) && return 0.0
+
+    aᵢ = sum(A, dims=2)
```
alyst: That looks cool, but I'm not sure it's immediately obvious from the var names that they refer to the cluster sizes of `a` and `b` from `mutualinfo(a, b)`.
FWIW, I'd keep it as it was in `vmeasure`.

wildart:
Should be fine now.

src/mutualinfo.jl
```diff
@@ -0,0 +1,40 @@
...
+"""
+    mutualinfo(a, b; normed=true) -> Float64
+
+Compute the *mutual information* between the two clusterings of the same
+data points.
+
+`a` and `b` can be either [`ClusteringResult`](@ref) instances or
+assignments vectors (`AbstractVector{<:Integer}`).
+
+If `normed` paremeter is `true`, the normalized mutual information (sum) is returned.
```
alyst:
```suggestion
If `normed` parameter is `true`, the normalized mutual information (sum) is returned.
```

alyst: I'm not sure I 100% understand "normalized mutual information (sum)" mean. What is "sum" in this context?

alyst:
@wildart Thank you! Could you please fix a tiny typo and maybe clarify a bit the "normed = true" case, then I'll merge the PR.

PR #168, opened 2019-05-21
wcounts() and tiny FuzzyCMeans API cleanups

alyst:
Some small counts/weights API improvements:
 * add `wcounts(ClusteringResult)` (sum of the point weights)
 * add `wcounts(FuzzyCMeansResult)` and `nclusters(FuzzyCMeansResult)`
 * `FuzzyCMeansResult` is not a `ClusteringResult` (cannot have counts or assignments)
 * rename `KMeansResult.cweights` into `KMeansResult.wcounts` for consistency (old property is still supported but deprecated)

PR #172, opened 2019-06-21
Switch from METADATA to registrator

kescobo:
With the [new package registration system](https://discourse.julialang.org/t/switching-package-registration-systems-soon/22677/144), REQUIRE is out and Project.toml is in.

This is an annoying but necessary task, so I thought I'd just do it. I used [this script](https://github.com/JuliaLang/Pkg.jl/pull/1133), and added a version (`0.14.0`, since I figured this warranted a major version bump).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,0 +1,29 @@
+name = "Clustering"
+uuid = "aaaa29a8-35af-508c-8bc3-b662a17a0fe5"
+
+[deps]
+Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+NearestNeighbors = "b8a86587-4115-5ab1-83bc-aa920d37bbce"
+Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
+
+[compat]
+Distances = "≥ 0.8.0"
+NearestNeighbors = "≥ 0.0.3"
+StatsBase = "≥ 0.9.0"
+julia = "≥ 1.0.0"
+
+[extras]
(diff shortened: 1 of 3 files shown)


Project.toml
```diff
@@ -0,0 +1,29 @@
...
+Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
+
+[compat]
+Distances = "≥ 0.8.0"
+NearestNeighbors = "≥ 0.0.3"
+StatsBase = "≥ 0.9.0"
+julia = "≥ 1.0.0"
```
nalimilan:
```suggestion
julia = "1"
```

nalimilan:
> `0.14.0`, since I figured this warranted a major version bump

Does it? I don't think so, as nothing changes at all for users (the registries already generated the Project.tool file automatically under the hood).

alyst:
Before the new release, I would like to merge #168, which is a (small) API change warranting a major version bump.
BTW, @nalimilan could you please take a look at #168?

kescobo:
>Does it? I don't think so, as nothing changes at all for users

Fair enough - I definitely don't feel strongly about it. Should I change to 0.13.2 then?

kescobo:
Fine by me either way. Let me know if you want me to change it or leave it alone :-)

alyst:
I think the consensus now is that the next release we will 0.13.2.
@kescobo Could you please commit the suggested change, so that I can merge the PR?

kescobo:
@alyst All set!

alyst:
@kescobo Thanks! Actually, I've never used Registrator before :) Should I add the comment `@JuliaRegistrator register + release notes` and then merge?

kescobo:
@alyst Evidently not? 😆 

I've only used BioJulia's registrator, but I did it in an issue after the PR was merged. Seems like it makes the most sense to do in in the PR once it's merged but ¯\_(ツ)_/¯

alyst:
Also I should be careful calling the registrator by name, since it doesn't understand the quotes. :)
I'll try commenting the head commit after merging the PR.

PR #174, opened 2019-08-21
bump version, up deps

matbesancon:


matbesancon:
> what was the specific motivation

While some (actually, all) of these dependencies are not in >=1.0, things may break between minor version changes, without intermediate deprecation. This is avoided but not guaranteed.

> this upgrade may unnecessary break the user code that relies on deprecated/removed StatsBase API

It would not break, if they rely on older StatsBase, it will just constraint Clustering to remain to the previous version

alyst:
> It would not break, if they rely on older StatsBase, it will just constraint Clustering to remain to the previous version

Hmm, I thought that with e.g. `StatsBase = 0.16` and in the absence of the other constraints *Clustering.jl* will automatically use the latest *StatsBase.jl* up to `1.0.0`.
But if the users require (in their *Project.toml*) `StatsBase = ~0.16` and *Clustering.jl* declares `StatsBase = 0.32`, there would be no versions that satisfy both dependencies.

kescobo:
>But if the users require (in their Project.toml) StatsBase = ~0.16 and Clustering.jl declares StatsBase = 0.32, there would be no versions that satisfy both dependencies.

This is only true if they put a version contraint on `Clustering.jl` that conflicts. If they just do `add Clustering StatsBase#v0.16`, it will install the latest version of Clustering.jl that works with that version of StatsBase. I think...

matbesancon:
yup that's how I understand it too. Better have a version incompatibility from Pkg than some weird error

nalimilan:
> yup that's how I understand it too. Better have a version incompatibility from Pkg than some weird error

Really, we've tried that and it's terribly painful. It's much better to fix version bounds in the (relatively unlikely) event that something breaks. Otherwise, we have to update all packages that depend on StatsBase each time we tag a minor release. I've done a huge PR like that once, and that was so messy that I won't do it again.

The only good solution is to reach 1.0.

matbesancon:
This also means testability is a pain, you can't guarantee things work without testing your dependencies with all versions from lower to upper bound

matbesancon:
ok, can we at least bump everything to its most current version and use the caret operator?

matbesancon:
@nalimilan check new version

nalimilan:
I'd rather leave everything as-is unless we know the stated versions don't work. Allowing older versions is useful precisely because some other packages require precise minor versions, and until they are updated that creates version conflicts.

Project.toml
```diff
@@ -12,9 +12,9 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
-Distances = "≥ 0.8.0"
-NearestNeighbors = "≥ 0.0.3"
-StatsBase = "≥ 0.9.0"
+Distances = "0.8 1"
```
nalimilan: This syntax isn't actually supported. I'm not sure it's possible to include all versions between 0.8 and 1 in the current system. Anyway it would be better to include 1.x, since it's unlikely that the move to 1.0 will be more breaking than the move from one minor version to the other (quite the opposite).

matbesancon: reverted thanks

matbesancon:
> Anyway it would be better to include 1.x, since it's unlikely that the move to 1.0 will be more breaking than the move from one minor version to the other (quite the opposite).

Not so sure, the move to 1.0 is the last breaking one you can do before 2.0

Project.toml
nalimilan: So we're back to the original issue... :-/

matbesancon:
Anyway as it is now it's basically not checking any upper bound so you could even let up to 2.0. I would be for restricting for peace of mind, some breaking bugs can be non-trivial to fix

matbesancon:
alright, I guess that's a nasty side-effect of the whole ecosystem being pre 1.0

alyst:
Just for my understanding (because I feel I cannot grasp the whole complexity of dependencies problem), what is wrong with changing from `Distances = ">= 0.8.0"` to `Distances = "0.8"`?
That would require fixing the compatibility for *Distances >= 1.x* later (but it's quite likely that something would be broken with major version bump anyway).
In case both 1.0 and pre-1.0 *Distances.jl* work fine with *Clustering.jl*, it should then be possible to specify `Distances = "0.8, 1"`, per *Pkg.jl* manual.

matbesancon:
> what is wrong with changing from Distances = ">= 0.8.0" to Distances = "0.8"?

It means that it will not use Distances 0.9 and above, since these could be breaking compared to 0.8. With `>= 0.8`, any version could be used, Distances 4.6 for instance, resulting in errors.

the >= constraint only restricts the oldest versions

nalimilan:
This is due to a special-casing of pre-1.0 versions in the compatibility rules: a minor release is considered as breaking there (in line with our conventions).

PR #185, opened 2020-02-25
kmeans corner cases for k=1 and k=n

### LINKED ISSUE #183: kmeans() and kmedoids() don't support k=1 and k=n
Right now if the number of centers `k` is equal to the number of points, `kmeans` returns an error, shouldn't the behavior be to return the whole data?

birm:
(possibly) closes #183 

It looks like kmedoids already supports k=1 and k=n, at least it doesn't throw an error if so.

So this PR only interacts with kmeans.

### DIFF
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -54,13 +54,27 @@ function kmeans!(X::AbstractMatrix{<:Real},                # in: data matrix (d 
     dc, k = size(centers)
 
     d == dc || throw(DimensionMismatch("Inconsistent array dimensions for `X` and `centers`."))
-    (2 <= k < n) || throw(ArgumentError("k must have 2 <= k < n=$n ($k given)."))
-    if weights !== nothing
-        length(weights) == n || throw(DimensionMismatch("Incorrect length of weights."))
-    end
+    (1 <= k <= n) || throw(ArgumentError("k must have 1 <= k <= n=$n ($k given)."))
+    if (k == 1 )
+      WC = (weights === nothing) ? Int : eltype(weights)
+      D = typeof(one(eltype(centers)) * one(WC))
+      # all data points belong to this one cluster
+      return KmeansResult(centers, Base.ones(Int, n), Base.zeros(D, n), Base.fill(n,1),
+                          Base.fill(n,1), D(1), 0, true)
+    elseif ( k == n)
+      WC = (weights === nothing) ? Int : eltype(weights)
+      D = typeof(one(eltype(centers)) * one(WC))
+      # each data point is its own cluster center
+      return KmeansResult(typeof(centers)(X), Base.cumsum(ones(Int, n)), Base.zeros(D, n), Base.ones(Int, k),
+                          Base.ones(Int, k), D(1), 0, true)
+    else
+      if weights !== nothing
+          length(weights) == n || throw(DimensionMismatch("Incorrect length of weights."))
+      end
 
-    _kmeans!(X, weights, centers, Int(maxiter), Float64(tol),
-             display_level(display), distance)
+      return _kmeans!(X, weights, centers, Int(maxiter), Float64(tol),
+               display_level(display), distance)
+    end
 end
 
 
@@ -92,7 +106,7 @@ function kmeans(X::AbstractMatrix{<:Real},                # in: data matrix (d x
                 display::Symbol=_kmeans_default_display,  # in: level of display
                 distance::SemiMetric=SqEuclidean())       # in: function to calculate distance with
     d, n = size(X)
-    (2 <= k < n) || throw(ArgumentError("k must be 2 <= k < n, k=$k given."))
+    (1 <= k <= n) || throw(ArgumentError("k must be 1 <= k <= n, k=$k given."))
 
     # initialize the centers using a type wide enough so that the updates
     # centers[i, cj] += X[i, j] * wj will occur without loss of precision through rounding
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -61,7 +61,7 @@ function kmeans!(X::AbstractMatrix{<:Real},                # in: data matrix (d 
       return KmeansResult(mean!(centers, X), fill(1, n), zeros(D, n), fill(n, 1),
                           fill(n, 1), D(1), 0, true)
     elseif k == n # each point in its own cluster
-      return KmeansResult(copyto!(centers, X), 1:k, zeros(D, k), fill(1, k),
+      return KmeansResult(copyto!(centers, X), collect(1:k), zeros(D, k), fill(1, k),
                           Base.ones(Int, k), D(1), 0, true)
     else
       if weights !== nothing
--- a/src/kmeans.jl
+++ b/src/kmeans.jl
@@ -52,20 +52,16 @@ function kmeans!(X::AbstractMatrix{<:Real},                # in: data matrix (d 
                  distance::SemiMetric=SqEuclidean())       # in: function to compute distances
     d, n = size(X)
     dc, k = size(centers)
+    WC = (weights === nothing) ? Int : eltype(weights)
+    D = typeof(one(eltype(centers)) * one(WC))
 
     d == dc || throw(DimensionMismatch("Inconsistent array dimensions for `X` and `centers`."))
-    (1 <= k <= n) || throw(ArgumentError("k must have 1 <= k <= n=$n ($k given)."))
-    if (k == 1 )
-      WC = (weights === nothing) ? Int : eltype(weights)
-      D = typeof(one(eltype(centers)) * one(WC))
-      # all data points belong to this one cluster
-      return KmeansResult(centers, Base.ones(Int, n), Base.zeros(D, n), Base.fill(n,1),
-                          Base.fill(n,1), D(1), 0, true)
-    elseif ( k == n)
-      WC = (weights === nothing) ? Int : eltype(weights)
-      D = typeof(one(eltype(centers)) * one(WC))
-      # each data point is its own cluster center
-      return KmeansResult(typeof(centers)(X), Base.cumsum(ones(Int, n)), Base.zeros(D, n), Base.ones(Int, k),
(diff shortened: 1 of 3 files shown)


src/kmeans.jl
```diff
@@ -54,13 +54,27 @@ function kmeans!(X::AbstractMatrix{<:Real},                # in: data matrix (d
     dc, k = size(centers)
 
     d == dc || throw(DimensionMismatch("Inconsistent array dimensions for `X` and `centers`."))
-    (2 <= k < n) || throw(ArgumentError("k must have 2 <= k < n=$n ($k given)."))
-    if weights !== nothing
-        length(weights) == n || throw(DimensionMismatch("Incorrect length of weights."))
-    end
+    (1 <= k <= n) || throw(ArgumentError("k must have 1 <= k <= n=$n ($k given)."))
```
alyst: Let's take the opportunity to improve the message:
```suggestion
    (1 <= k <= n) || throw(ArgumentError("k must be from 1:n (n=$n), k=$k given."))
```

src/kmeans.jl
```diff
@@ -92,7 +106,7 @@ function kmeans(X::AbstractMatrix{<:Real},                # in: data matrix (d x
                 display::Symbol=_kmeans_default_display,  # in: level of display
                 distance::SemiMetric=SqEuclidean())       # in: function to calculate distance with
     d, n = size(X)
-    (2 <= k < n) || throw(ArgumentError("k must be 2 <= k < n, k=$k given."))
+    (1 <= k <= n) || throw(ArgumentError("k must be 1 <= k <= n, k=$k given."))
```
```suggestion
    (1 <= k <= n) || throw(ArgumentError("k must be from 1:n (n=$n), k=$k given."))
```

src/kmeans.jl
```diff
@@ -54,13 +54,27 @@ function kmeans!(X::AbstractMatrix{<:Real},                # in: data matrix (d
...
+                          Base.ones(Int, k), D(1), 0, true)
+    else
+      if weights !== nothing
+          length(weights) == n || throw(DimensionMismatch("Incorrect length of weights."))
+      end
 
-    _kmeans!(X, weights, centers, Int(maxiter), Float64(tol),
-             display_level(display), distance)
+      return _kmeans!(X, weights, centers, Int(maxiter), Float64(tol),
+               display_level(display), distance)
```
alyst: Pls fix the indentation

src/kmeans.jl
```diff
@@ -54,13 +54,27 @@ function kmeans!(X::AbstractMatrix{<:Real},                # in: data matrix (d
...
+      WC = (weights === nothing) ? Int : eltype(weights)
+      D = typeof(one(eltype(centers)) * one(WC))
+      # all data points belong to this one cluster
+      return KmeansResult(centers, Base.ones(Int, n), Base.zeros(D, n), Base.fill(n,1),
+                          Base.fill(n,1), D(1), 0, true)
+    elseif ( k == n)
+      WC = (weights === nothing) ? Int : eltype(weights)
+      D = typeof(one(eltype(centers)) * one(WC))
+      # each data point is its own cluster center
+      return KmeansResult(typeof(centers)(X), Base.cumsum(ones(Int, n)), Base.zeros(D, n), Base.ones(Int, k),
```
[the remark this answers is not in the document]
birm: You're right; collecting does seem to fit better. (It was necessary to collect, by the way)

birm:
Thanks for the suggestions. I'm not sure if the kmedoids tests are any good, so let me know if you have a better idea on how to test that functionality.

test/kmeans.jl
```diff
@@ -32,6 +32,15 @@ end
     end
 end
 
+@testset "k=1 and k=n corner cases" begin
+    Random.seed!(34568)
+    @test length(kmeans(randn(2,3), 1).centers) == 2
+    @test length(kmeans(randn(2,3), 3).centers) == 6
```
alyst: I think we don't really need the randomized examples here, since we are checking the fixed one.

test/kmeans.jl
```diff
@@ -32,6 +32,15 @@ end
...
+    x = [[ 0.5 1 ]; [ 1 0.5 ]]
```
alyst: Simpler form:
```suggestion
    x = [0.5 1; 1 0.5]
```

What about 3-point example? I think we have more chances to detect when something is wrong with 3 points:
```suggestion
    x = [0.5 1 2; 1 0.5 0]
```

test/kmeans.jl
```diff
@@ -32,6 +32,15 @@ end
...
+    @test dropdims(kmeans(x,1).centers, dims=2) == [0.75, 0.75]
```
alyst: Since we have already checked (in the previous testset) that `kmeans()` returns KmeansResults, we can do it like
```suggestion
    kmeans1 = kmeans(x, 1)
    @test kmeans1.centers == reshape([0.75, 0.75], (2, 1))
```
You need also to test that the other fields are set correctly (*counts*, *assignments*, *wcounts*).
Also you need to add the weighted version test as well and check that the weights are actually used (*wcounts* should be different than the unweighted version).

test/kmeans.jl
```diff
@@ -32,6 +32,15 @@ end
...
+    @test kmeans(x,2).centers == x
```
alyst: Same as comment above, we need to test the other fields as well.

test/kmedoids.jl
```diff
@@ -39,6 +39,11 @@ R = kmedoids(costs, k)
 @test isapprox(sum(R.acosts), R.totalcost)
 @test R.converged
 
+# k=1 and k=n cases
+x = pairwise(SqEuclidean(), [[ 1 2 3]; [.1 .2 .3]; [4 5.6 7]], dims=2)
```
alyst:
```suggestion
x = pairwise(SqEuclidean(), [1 2 3; .1 .2 .3; 4 5.6 7], dims=2)
```

test/kmedoids.jl
```diff
@@ -39,6 +39,11 @@ R = kmedoids(costs, k)
...
+@test nclusters(kmedoids(x, 1)) == 1
+@test nclusters(kmedoids(x, 3)) ==  3
```
alyst:
```suggestion
kmed3 = kmedoids(x, 3)
@test nclusters(kmed3) == 3
@test assignments(kmed3) == [1, 2, 3]
@test kmed3.medoids == [1, 2, 3]
```

test/kmedoids.jl
```diff
@@ -39,6 +39,11 @@ R = kmedoids(costs, k)
 @test isapprox(sum(R.acosts), R.totalcost)
 @test R.converged
 
+# k=1 and k=n cases
+x = pairwise(SqEuclidean(), [[ 1 2 3]; [.1 .2 .3]; [4 5.6 7]], dims=2)
+@test nclusters(kmedoids(x, 1)) == 1
```
alyst: You may need to tweak the coordinates that you use to generate `x` distance matrix, so that the choice of medoid point is more obvious, and that its index is different from 1 (the last test would need to be fixed then).
```suggestion
kmed1 = kmedoids(x, 1)
@test nclusters(kmed1) == 1
@test assignments(kmed1) == [1, 1, 1]
@test kmed1.medoids == [1]
```

test/kmedoids.jl
birm: This isn't reliable:
```
julia> kmed3 = kmedoids(x, 3).medoids
3-element Array{Int64,1}:
 1
 3
 2

julia> kmed3 = kmedoids(x, 3).medoids
3-element Array{Int64,1}:
 3
 1
 2

julia> kmed3 = kmedoids(x, 3).medoids
3-element Array{Int64,1}:
 3
 1
 2

julia> kmed3 = kmedoids(x, 3).medoids
3-element Array{Int64,1}:
 3
 2
 1
```

alyst: One way to fix it is
```julia
@test sort(kmed3.medoids) == [1, 2, 3] # the order depends on the initial random seeds
```

test/kmeans.jl
birm: Looking at it, the _old_ version also accepts k=2 in a 2x? matrix input; I'll use a 3x3 instead.

alyst: Hmm, that's strange. The columns are the datapoints, so it should throw for *k=2* and 2x1 or 2x2 *x* matrix.

test/kmedoids.jl
birm: I hope that use of Set was an acceptable alternative :)

birm:
Thanks for the suggestions and helping with the weighted case!

test/kmeans.jl
```diff
@@ -32,6 +32,21 @@ end
...
+    x = [0.5 1 2; 1 0.5 0; 3 2 1]
+    km1 = kmeans(x,1)
+    @test km1.centers == reshape([7. /6,0.5, 2.0], (3,1))
+    @test km1.counts == fill(3, (1))
+    @test km1.assignments == fill(1, (3))
+    km3 = kmeans(x,3)
+    @test km3.centers == x
+    @test km3.counts == fill(1, (3))
+    @test km3.assignments == collect(1:3)
+    w = Vector{Int}([1,2,3])
```
alyst: I'd test with something different than `Int`, because it's the default type when no weights are specified.
Also use something different than *.assignments*:
```suggestion
    w = [0.5, 2.0, 1.0]
```

birm:
The kmedoids k=1 test case is less straightforward. Since we chose x, it's consistent, but the behavior is not generally so:

```

julia> kmedoids(pairwise(SqEuclidean(),rand(3,3), dims=2),1).medoids
1-element Array{Int64,1}:
 3

julia> kmedoids(pairwise(SqEuclidean(),rand(3,3), dims=2),1).medoids
1-element Array{Int64,1}:
 3

julia> kmedoids(pairwise(SqEuclidean(),rand(3,3), dims=2),1).medoids
1-element Array{Int64,1}:
 2

julia> kmedoids(pairwise(SqEuclidean(),rand(3,3), dims=2),1).medoids
1-element Array{Int64,1}:
 2
```

PR #186, opened 2020-02-26
rename kmedoids acost to cost re #98

birm:
closes #98 

I figured it made sense to provide a finished version of #98, since it was seen to have been a good idea at the time.

However, it'll quite likely break some workflows. Is it still desired?

### DIFF
--- a/src/kmedoids.jl
+++ b/src/kmedoids.jl
@@ -56,7 +56,7 @@ const _kmed_default_display = :none
     kmedoids(dist::DenseMatrix, k::Integer; ...) -> KmedoidsResult
 
 Perform K-medoids clustering of ``n`` points into `k` clusters,
-given the `dist` matrix (``n×n``, `dist[i, j]` is the cost of
+given the `distance` matrix (``n×n``, `dist[i, j]` is the cost of
 assigning `j`-th point to the medoid represented by the `i`-th point).
 
 # Note
@@ -123,7 +123,7 @@ end
 #### core algorithm
 
 function _kmedoids!(medoids::Vector{Int},      # initialized medoids
-                    dist::DenseMatrix{T},      # cost matrix
+                    dist::DenseMatrix{T},      # distance matrix
                     maxiter::Int,              # maximum number of iterations
                     tol::Real,                 # tolerable change of objective
                     displevel::Int) where T<:Real            # level of display
--- a/src/kmedoids.jl
+++ b/src/kmedoids.jl
@@ -32,11 +32,11 @@ end
 
 # FIXME remove after deprecation period for acosts
 Base.propertynames(kmed::KmedoidsResult, private::Bool = false) =
-    (fieldnames(kmed)..., #= deprecated =# :acosts)
+    (fieldnames(kmed)..., #= deprecated since v0.13.4=# :acosts)
 
 # FIXME remove after deprecation period for acosts
 function Base.getproperty(kmed::KmedoidsResult, prop::Symbol)
-    if prop == :acosts
+    if prop == :acosts # deprecated since v0.13.4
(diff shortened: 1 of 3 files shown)


alyst:
I think it makes sense.
Actually it's possible to support *.acosts* in deprecated mode, see [how it was done for Hclust](https://github.com/JuliaStats/Clustering.jl/commit/010d408759eb38c0bfc9751d512a89fe2d9a66e1).

src/kmedoids.jl
```diff
@@ -13,7 +13,7 @@ The output of [`kmedoids`](@ref) function.
  - `assignments::Vector{Int}`: the indices of clusters the points are assigned
    to, so that `medoids[assignments[i]]` is the index of the medoid for the
    ``i``-th point
- - `acosts::Vector{T}`: assignment costs, i.e. `acosts[i]` is the cost of
+ - `costs::Vector{T}`: assignment costs, i.e. `acosts[i]` is the cost of
```
alyst:
```suggestion
 - `costs::Vector{T}`: assignment costs, i.e. `costs[i]` is the cost of
```
Could you check that all references of `acosts` are fixed? E.g. in docs?

src/kmedoids.jl
```diff
@@ -23,13 +23,28 @@ The output of [`kmedoids`](@ref) function.
...
     counts::Vector{Int}         # number of points assigned to each cluster (k)
     totalcost::Float64          # total assignment cost (i.e. objective) (k)
     iterations::Int             # number of elapsed iterations
     converged::Bool             # whether the procedure converged
 end
 
+# FIXME remove after deprecation period for acosts
+Base.propertynames(kmed::KmedoidsResult, private::Bool = false) =
+    (:medoids, :assignments, :costs, :counts, :totalcost, :iterations, :converged,
+     #= deprecated =# :acosts)
```
alyst: I think it was not available when I was writing *Hclust* deprecations, but now one can write
```suggestion
    (fieldnames(kmed)..., #= deprecated =# :acosts)
```
which would allow modifying `KmedoidsResult` without touching this method

src/kmedoids.jl
```diff
@@ -23,13 +23,28 @@ The output of [`kmedoids`](@ref) function.
 mutable struct KmedoidsResult{T} <: ClusteringResult
     medoids::Vector{Int}        # indices of methods (k)
     assignments::Vector{Int}    # assignments (n)
-    acosts::Vector{T}           # costs of the resultant assignments (n)
+    costs::Vector{T}           # costs of the resultant assignments (n)
```
alyst:
```suggestion
    costs::Vector{T}            # costs of the resultant assignments (n)
```

src/kmedoids.jl
```diff
@@ -13,23 +13,37 @@ The output of [`kmedoids`](@ref) function.
...
+    costs::Vector{T}            # costs of the resultant assignments (n)
     counts::Vector{Int}         # number of points assigned to each cluster (k)
     totalcost::Float64          # total assignment cost (i.e. objective) (k)
     iterations::Int             # number of elapsed iterations
     converged::Bool             # whether the procedure converged
 end
 
+# FIXME remove after deprecation period for acosts
+Base.propertynames(kmed::KmedoidsResult, private::Bool = false) =
+    (fieldnames(kmed)..., #= deprecated =# :acosts)
```
alyst:
```suggestion
    (fieldnames(kmed)..., #= deprecated since v0.13.4=# :acosts)
```

src/kmedoids.jl
```diff
@@ -13,23 +13,37 @@ The output of [`kmedoids`](@ref) function.
...
+
+# FIXME remove after deprecation period for acosts
+function Base.getproperty(kmed::KmedoidsResult, prop::Symbol)
+    if prop == :acosts
```
alyst:
```suggestion
    if prop == :acosts # deprecated since v0.13.4
```

PR #193, opened 2020-03-07
Make accept AbstractMatrix

oxinabox:
Affprop, DBScan and K-Mediods were dispatching on `DenseMatrix`.
This is a uncomfortable compromise as a great many matrix types, especially wrapper types, subtype `AbstractMatrix`.
(This makes sense the wrapper type doesn't know what it is wrapping).

I would be kind a surprised if any of the algorithms didn't work on structured or sparse matrixes.
Albeit with reduced performance and possibly large allocations.
But if they don't work for a particular type then that will throw its own error -- we don't need to  `MethodError` right at the start.

Examples of `AbstractMatrix` this should now support `Transpose`,`PermuteDims` and `SubArray`, `Symmetric` (rather makes sense for distances). And things from packages like `NamedDimsArray` and `StaticArray`.

I have added tests with the distance matrixes as `SubArray` and `Symmetric`, to make sure it works.

### DIFF
--- a/test/affprop.jl
+++ b/test/affprop.jl
@@ -6,6 +6,7 @@ using Clustering
 using LinearAlgebra
 using Random
 using Statistics
+include("test_helpers.jl")
 
 @testset "affinityprop() (affinity propagation)" begin
     @testset "Argument checks" begin
@@ -47,11 +48,10 @@ using Statistics
     end
 
     @testset "Support for arrays other than Matrix{T}" begin
-        R2 = affinityprop(@view S[:,:])  # run on complete subarray
-        @test R2.assignments == R.assignments
-
-        R3 = affinityprop(Symmetric(S))  # run on complete subarray
-        @test R3.assignments == R.assignments
+        @testset "$(typeof(M))" for M in equivalent_matrices(S)
+            R2 = affinityprop(M)
+            @test R2.assignments == R.assignments
+        end
     end
 
     #= compare with python result
--- a/test/affprop.jl
+++ b/test/affprop.jl
@@ -46,6 +46,11 @@ using Statistics
         @test R.counts[i] == count(==(i), R.assignments)
     end
 
+    @testset "Ensure works on nonbasic array type (SubArray)" begin
+        RR = affinityprop(@view S[:,:])  # run on complete subarray
+        @test RR.assignments == R.assignments
+    end
+
     #= compare with python result
     the reference assignments were computed using python sklearn:
     ```julia
--- a/test/affprop.jl
+++ b/test/affprop.jl
@@ -46,7 +46,7 @@ using Statistics
         @test R.counts[i] == count(==(i), R.assignments)
     end
 
-    @testset "Ensure works on nonbasic array type" begin
+    @testset "Support for arrays other than Matrix{T}" begin
         R2 = affinityprop(@view S[:,:])  # run on complete subarray
         @test R2.assignments == R.assignments
 
--- a/test/affprop.jl
+++ b/test/affprop.jl
@@ -46,9 +46,12 @@ using Statistics
         @test R.counts[i] == count(==(i), R.assignments)
     end
 
-    @testset "Ensure works on nonbasic array type (SubArray)" begin
-        RR = affinityprop(@view S[:,:])  # run on complete subarray
-        @test RR.assignments == R.assignments
+    @testset "Ensure works on nonbasic array type" begin
+        R2 = affinityprop(@view S[:,:])  # run on complete subarray
+        @test R2.assignments == R.assignments
+
+        R3 = affinityprop(Symmetric(S))  # run on complete subarray
+        @test R3.assignments == R.assignments
     end
 
     #= compare with python result
--- a/test/test_helpers.jl
+++ b/test/test_helpers.jl
@@ -1,0 +1,24 @@
+using LinearAlgebra
+using SparseArrays
+
+"""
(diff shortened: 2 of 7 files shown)


oxinabox:
Bump

test/affprop.jl
```diff
@@ -46,6 +46,14 @@ using Statistics
         @test R.counts[i] == count(==(i), R.assignments)
     end
 
+    @testset "Ensure works on nonbasic array type" begin
```
alyst: I think it's slightly more informative:
```suggestion
    @testset "non-Matrix{T} distances support" begin
```
Could you please adjust the other testset names as well?

test/affprop.jl
```diff
@@ -46,6 +46,14 @@ using Statistics
...
+        R2 = affinityprop(@view S[:,:])  # run on complete subarray
```
alyst: It's as long as the previous version, but a bit more explicit (and without macro magic ;)).
Could you please update it in the other tests as well?
```suggestion
        R2 = affinityprop(view(S, 1:n, 1:n))  # run on complete subarray
```

oxinabox: That is a strictly worse change.

`@view S[:, :]` is the same as `view(S, :, :)`. *not* the same as `@view(S, 1:n, 1:n)`
Putting in the `1:n` makes it hit different a different code path.
Now it has to create a `Base.OneTo(n)` object and do a bounds-check.
So it is slower.
```
julia> using BenchmarkTools

julia> const d = rand(1000,1000);

julia> const n = size(d,1);

julia> @btime(@view ($d)[:,:]);
  6.964 ns (1 allocation: 48 bytes)

julia> @btime(view($d, 1:$n, 1:$n));
  8.807 ns (1 allocation: 64 bytes)

julia> @btime(view($d, :, :));
  6.875 ns (1 allocation: 48 bytes)
```

Its also more confusing to read since I need to go and find what `n` was defined to be.
Where as `:` means what it always means; the whole dimension.

Further more using `1:n` breaks for OffsetArrays etc.
so its a bad pattern to follow.

alyst: > That is a strictly worse change.

Obviously, I don't recommend using it in the actual code.
The intention is to create `AbstractMatrix` that is more different from `Matrix` than `view(M, :, :)`, so that -- if the clustering method does not fully support `AbstractArray` in some way -- we have a bit more chances to detect it.
From that perspective `view(M, row_ixs, col_ixs)` or `sparse()` would be even better, since it would test whether the clustering method relies on stride.

But I'm not bound to this suggestion, especially if `1:n` hits the performance of some tests hardly.
I guess we will be adding some type checks/warnings as the users will discover the limitations of `::AbstractMatrix`, so it's fine even with these very basic tests.

oxinabox: > hits the performance of some tests hardly.

It's not about performance here, it is about being consistent with the code you would write else where. And everywhere you would use `:` since in other places performance *does* mater.

> IMHO using @view does make sense to convert multiple indexing operations at once or to turn views on/off, otherwise it looks confusing.

`@view` does not do that. You are thinking of `@views` which applies to a single block of code.
`@view` only applies to a single indexing expression, not to a block of code.
`@view` makes sure that writing `xs[5:end]` works, since can't do `@view

---

> From that perspective view(M, row_ixs, col_ixs) or sparse() would be even better, since it would test whether the clustering method relies on stride.

If that is your goal then we can't use `view(x, 1:n, 1:n)` either since `strides` works fine on that. we will need to densify the indexed.
Which can be written clearest as `@view x[collect(1:end), collect(1:end)]`
but since I can't use the macro so.

alyst: > @view does not do that. You are thinking of @views which applies to a single block of code.

My bad, thanks for clarification.

I hope we understood each other.
If you want to go deeper with testing various `AbstractMatrix` corner cases, you are welcome to do so.
I was just pointing you to this direction with my original comment on `view(S, 1:n, 1:n)`, but the possibilities are endless, and I'm fine with `view(S, :, :)` as a starting point.

But, please, change to `view(S, ...)` instead of `@view` macro.
Yes, it could be regarded as a matter of personal preference, but it's also about the code consistency, since in the other parts of the package we use `view()`.

Thank you!

test/affprop.jl
```diff
@@ -6,6 +6,7 @@ using Clustering
 using LinearAlgebra
 using Random
 using Statistics
+include("test_helpers.jl")
```
oxinabox: Do we like reincluding this in all the files? (rather than in runtests.jl)
It means we can run each file seperately, which I guess is the reason all the `using`s are repeated.
So I guess it is right to follow the pattern :-) ?

test/test_helpers.jl
```diff
@@ -0,0 +1,24 @@
+using LinearAlgebra
+using SparseArrays
+
+"""
+    equivalent_matrices(x::AbstractMatrix)
+
+Returns a collection of matrixes that are equal to the input `x`, but of a different type.
+Useful for testing if things still work on different types of matrix.
+"""
+function equivalent_matrices(x::AbstractMatrix)
```
oxinabox: If we wanted to go more intense with this we could nest if so we would see things like `view(sparse(x), :, :)`
etc.

oxinabox:
bump, this should be ready to go

test/affprop.jl
oxinabox: this is done

wildart:
LGTM

alyst:
Thanks!

PR #208, opened 2021-01-28
Update fuzzycmeans.md

alexpattyn:
Added an example about fuzzy c means on 2D images.


alexpattyn:
Perhaps double check my comment: 
```
This image represents how much each pixel 
belongs to a given segment. For example here we 
are looking at segment 1. 
```

^not sure if that is fully accurate.

PR #220, opened 2021-08-22
Added pair counting fmeasure metric

dinarior:
I often use this metric, I think it's worth having.

refs:
https://nlp.stanford.edu/IR-book/html/htmledition/evaluation-of-clustering-1.html

https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.214.7233&rep=rep1&type=pdf

Included also is precision and recall for clustering, I was not sure about the proper name (e.g. precision is already in use by Julia base).

The _pair_confusion_matrix is translated from sklearn's https://github.com/scikit-learn/scikit-learn/blob/2beed55847ee70d363bdbfe14ee4401438fba057/sklearn/metrics/cluster/_supervised.py#L154

there is a small duplication with the rand index, which also require this matrix, but as I did not want to modify it to use my new function.

### DIFF
--- a/src/Clustering.jl
+++ b/src/Clustering.jl
@@ -64,7 +64,10 @@ module Clustering
     Hclust, hclust, cutree,
 
     # MCL
-    mcl, MCLResult
+    mcl, MCLResult,
+
+    #fmeasure
+    pair_precision, fmeasure, pair_recall
 
     ## source files
 
@@ -84,6 +87,7 @@ module Clustering
     include("varinfo.jl")
     include("vmeasure.jl")
     include("mutualinfo.jl")
+    include("fmeasure.jl")
 
     include("hclust.jl")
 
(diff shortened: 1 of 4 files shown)


nalimilan:
Given how short and simple the code is, it probably won't have to be considered as derived from NumPy *if* you adapt it to make it more Julian and more efficient, as in the end the only think that will remain from NumPy is the algorithm. For example, `sum(c.*c)` should we written as `sum(abs2, c)`, `sum(c,dims=1)[:]` as `vec(sum(c, dims=1))` and so on.

dinarior:
Thanks,
I intend to rewrite it, maybe extract common functionalities from the ARI metric, hopefully, will get to it soon enough.

wildart:
I just re-implemented this functionality in #227 to fix ARI calculations.

dinarior:
Great!, I will wait for it to get pushed and update this commit accordingly.

PR #224, opened 2021-12-16
Disambiguate use of "pairwise" function

MartaVanin:
Hi, 
I am working on a package that depends on clustering.jl and got the following error while trying to perform kmeans clustering.
image
In this PR, I just add `Distances.` in front of all the `pairwise`s, so that it becomes clear.


wildart:
It should not be a problem now as both `Distances` and `StatsBase` extend `pairwise` from `StatsAPI`. Try to update your project.

MartaVanin:
thanks, updating seems to work and I can't reproduce the error anymore, so I am closing this

Issue #226, opened 2021-12-23
Adjusted Rand Index inconsistency with Python's sklearn implementation


jindravo:
Hello!

I've been working on a model that clusters MoGs into clusters and noticed a weird drop in performance when all MoGs in given input belong to only one cluster. After doing a little bit of sleuthing, I found out that there is an inconsistency between the implementation of ARI in `Clustering.jl` and `sklearn`. Specifically, the special case is handled [here in `Clustering.jl`](https://github.com/JuliaStats/Clustering.jl/blob/master/src/randindex.jl#L39) and [here in `sklearn`](https://github.com/scikit-learn/scikit-learn/blob/0d378913b/sklearn/metrics/cluster/_supervised.py#L396), where `Clustering.jl` assigns `ARI = 0` while `sklearn` assigns `ARI = 1`.

Intuitively, when all points belong to one cluster and the model correctly predicts that, one would expect ARI to be 1. Here follows the code to reproduce the scenario:

```jl
import Clustering: randindex

randindex([1, 1, 1], [1, 1, 1]) == 1.0 # false
```

```py
from sklearn.metrics import adjusted_rand_score as ari

ari([1, 1, 1], [1, 1, 1]) == 1.0 # True
```

Moreover, while debugging, I came across another case where the result differs, and although it's definitely not as crucial as it rarely happens in real life scenarios, I'll drop it here as well. Whenever each cluster has exactly one point belonging to it, the special case condition will also set the ARI to 0, even though one would expect 1 since the clustering is perfect.

```jl
import Clustering: randindex

randindex([1, 2, 3], [1, 2, 3]) == 1.0 # false
```

```py
from sklearn.metrics import adjusted_rand_score as ari

ari([1, 2, 3], [1, 2, 3]) == 1.0 # True
```

I'm assuming this in not the expected behaviour, hence I'm submitting an issue, however if it is, I'll be happy to hear an explanation with hopes that it'll change the intuition I see behind ARI.

jindravo:
Thanks a lot @wildart!

PR #227, opened 2021-12-25
fixed ARI (fixes #225 & #226)

### LINKED ISSUE #225: Adjusted Rand index inconsistency for large n
The adjusted Rand index fails unexpectedly when n is large (n > 100,000). Here is an example with a comparison to an R implementation.
```
using Random
using Clustering
using RCall
Random.seed!(123);

n = 100_000;
a = rand(1:3,n);
b = rand(1:3,n);

randindex(a,b)[1]

only(R"library(mclust); adjustedRandIndex($a,$b)")
```
which gives
```
0.2933142400616828

-1.5731751561282826e-6
```
In theory the true adjusted Rand index should be close to 0. This starts to happen around `n=83,000` for me.

As a Julia comparison, my own implementation of the adjusted Rand index gives the same result as in R:
```
function ari(a,b)
    table = counts(a,b)
    acounts = sum(table,dims=1)
    bcounts = sum(table,dims=2)
    
    score = sum([x*(x-1)/2 for x in table])
    asum = sum([x*(x-1)/2 for x in acounts])
    bsum = sum([x*(x-1)/2 for x in bcounts])
    expected = asum*bsum/binomial(sum(table),2)
    total = (asum + bsum)/2
    
    if total == expected
        return 0
    else
        return (score-expected)/(total-expected)
    end
end;
ari(a,b)
```
```
-1.5731751561282826e-6
```
I use Clustering.jl 0.14.2, Julia 1.6.2.

wildart:
- added pair confusion matrix (CM) calculations
- fixed ARI calculation using CM

### DIFF
--- a/src/confusion.jl
+++ b/src/confusion.jl
@@ -28,9 +28,7 @@ function confusion(a::AbstractVector{<:Integer}, b::AbstractVector{<:Integer})
 
     t2 = sum(abs2, c)                      # sum over rows & columns of nij^2
     t3 = nis+njs
-    #println("n: $n")
     C = Int[(t2-n)/2 (nis-t2)/2; (njs-t2)/2 (t2+n^2-t3)/2]
-    #println("C: $C")
     return C
 end
 confusion(a::ClusteringResult, b::ClusteringResult) =
--- a/src/confusion.jl
+++ b/src/confusion.jl
@@ -1,0 +1,42 @@
+"""
+    confusion(a::ClusteringResult, b::ClusteringResult) -> Matrix{Int}
+    confusion(a::ClusteringResult, b::AbstractVector{<:Integer}) -> Matrix{Int}
+    confusion(a::AbstractVector{<:Integer}, b::ClusteringResult) -> Matrix{Int}
+    confusion(a::AbstractVector{<:Integer}, b::AbstractVector{<:Integer}) -> Matrix{Int}
+
+Return 2x2 confusion matrix `C` that represents partition co-occurrence or
+similarity matrix between two clusterings by considering all pairs of samples
+and counting  pairs that are assigned into the same or into different clusters
+under the true and predicted clusterings.
+
+Considering a pair of samples that is in the same group as a **positive pair**,
+and a pair is in the different group as a **negative pair**, then the count of
+true positives is `C₀₀`, false negatives is `C₀₁`, false positives `C₁₀`, and
+true negatives is `C₁₁`:
+
+|  | Positive | Negative |
+|:--:|:-:|:-:|
+|Positive|C₀₀|C₁₀|
+|Negative|C₀₁|C₁₁|
+"""
+function confusion(a::AbstractVector{<:Integer}, b::AbstractVector{<:Integer})
+    c = counts(a, b)
+
+    n = sum(c)
+    nis = sum(abs2, sum(c, dims=2))        # sum of squares of sums of rows
+    njs = sum(abs2, sum(c, dims=1))        # sum of squares of sums of columns
+
+    t2 = sum(abs2, c)                      # sum over rows & columns of nij^2
+    t3 = nis+njs
+    #println("n: $n")
+    C = Int[(t2-n)/2 (nis-t2)/2; (njs-t2)/2 (t2+n^2-t3)/2]
+    #println("C: $C")
+    return C
+end
+confusion(a::ClusteringResult, b::ClusteringResult) =
+    confusion(assignments(a), assignments(b))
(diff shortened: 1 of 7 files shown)


wildart:
As `rand` behavior changed in recent julia versions, the rework of RNG handling or a test review is required.

pcjentsch:
This PR passes the relevant tests, can it be merged?

This is an old bug that has had other PRs to fix it: https://github.com/JuliaStats/Clustering.jl/pull/99

ararslan:
Looks like tests are passing except for 32-bit Linux. I wonder if the forced conversion to `Int` in `confusion` means that something in `randindex` overflows when `Int` is only half as big.

test/randindex.jl
```diff
@@ -34,4 +34,9 @@ a3 = [3, 3, 3, 2, 2, 2, 1, 1, 1, 1]
 
 @test randindex(a1, a2) == randindex(a2, a1)
 
+@test randindex(ones(Int, 3), ones(Int, 3)) == (1, 1, 0, 1)
+
+a,b = rand(1:5,100_000), rand(1:5,100_000)
```
pcjentsch, line 39: Yes the problem is this test is too big for 32 bit systems. We can just remove a zero from the length of these vectors and it's fine.

```
a,b = rand(1:5,10000), rand(1:5,10000)
```

I'm not sure if it's worth it to add a check in `confusion`.

pcjentsch:
Hm, not sure why that change broke the test. Sorry about that, I did test it locally before suggesting it.

src/confusion.jl
```diff
@@ -0,0 +1,40 @@
...
+    confusion(a::AbstractVector{<:Integer}, b::AbstractVector{<:Integer}) -> Matrix{Int}
+
+Return 2x2 confusion matrix `C` that represents partition co-occurrence or
+similarity matrix between two clusterings by considering all pairs of samples
+and counting  pairs that are assigned into the same or into different clusters
+under the true and predicted clusterings.
+
+Considering a pair of samples that is in the same group as a **positive pair**,
+and a pair is in the different group as a **negative pair**, then the count of
+true positives is `C₀₀`, false negatives is `C₀₁`, false positives `C₁₀`, and
```
alyst, line 14: Could you please update it to 1-based indexing of the matrix `C`?

src/randindex.jl
```diff
@@ -18,35 +18,25 @@ Returns a tuple of indices:
...
-    nis = sum(abs2, sum(c, dims=2))        # sum of squares of sums of rows
-    njs = sum(abs2, sum(c, dims=1))        # sum of squares of sums of columns
-
-    t1 = binomial(n, 2)                    # total number of pairs of entities
-    t2 = sum(abs2, c)                      # sum over rows & columnns of nij^2
-    t3 = .5*(nis+njs)
-
-    # Expected index (for adjustment)
-    nc = (n*(n^2+1)-(n+1)*nis-(n+1)*njs+2*(nis*njs)/n)/(2*(n-1))
+    a, c, b, d = confusion(a,b) # Table 2 from Steinley 2004
```
alyst, line 26:
```suggestion
    c11, c21, c12, c22 = confusion(a, b) # Table 2 from Steinley 2004
```

src/randindex.jl
```diff
@@ -18,35 +18,25 @@ Returns a tuple of indices:
...
 
-    A = t1+t2-t3;        # agreements count
-    D = -t2+t3;          # disagreements count
+    t = a+ b + c + d   # total number of pairs of entities
+    A = a + d
+    D = b + c
```
alyst, line 30:
```suggestion
    t = c12 + c21 + c12 + c22   # total number of pairs of entities
    A = c11 + c22
    D = c12 + c21
```

test/randindex.jl
```diff
@@ -34,4 +34,9 @@ a3 = [3, 3, 3, 2, 2, 2, 1, 1, 1, 1]
 
 @test randindex(a1, a2) == randindex(a2, a1)
 
+@test randindex(ones(Int, 3), ones(Int, 3)) == (1, 1, 0, 1)
+
+a,b = rand(1:5,10_000), rand(1:5,10_000)
+@test randindex(a,b)[1] < 1.0e-2
```
alyst, line 40:
```suggestion
a, b = rand(1:5, 10_000), rand(1:5, 10_000)
@test randindex(a, b)[1] < 1.0e-2
```

alyst (commented):
Thanks for the PR!
Looks good. Could you please implement the suggestions (which are basically just formatting improvements)?

alyst:
@wildart So sorry I overlooked your updates! Thank you!

PR #228, opened 2021-12-25
Add `rng` parameter to appropriate methods

wildart:
This PR adds a kw-argument `rng` to the methods that rely on RNG. This will break dependency on the default global RNG object and will allow to better control of RNG behavior in the tests by using RNG from `StableRNGs` package.

In addition, some test were switched from using `Random.GLOBAL_RNG` to `StableRNG`.


ararslan (approved):

adrhill:
Awesome, this looks great for tests in downstream packages. I would love to see a release tagged for this!

PR #236, opened 2022-10-11
Update dbscan.jl to support Arrow.Table files

ma-seefelder:
Dear all, 
I recently tried to perform clustering with DBSCAN algorithm. Unfortunately, my distance matrix and the raw data matrix are too large to fit into memory (> 100 GB). Hence, I usually save the data in batches using the Julia Package Arrow.jl, since it allows to memory map huge the data and fast indexing. Hence, I changed the functions for DBSCAN clustering to be compatible with the Arrow.Table data format. I thought that the support of the Arrow.Table format might interest also other people working with huge datasets. 

If you are interested in adding this functionality, I can also update the documentation or add the support to other clustering algortihms. 

Best wishes.

### DIFF
--- a/src/dbscan.jl
+++ b/src/dbscan.jl
@@ -70,7 +70,7 @@ function dbscan(D::Arrow.Table, ϵ::Real, minpts::Int; T = Base.Float64)
     # check arguments    
     eps > 0 || throw(ArgumentError("ϵ must be a positive value ($eps given)."))
     minpts >= 1 || throw(ArgumentError("minpts must be positive integer ($minpts given)."))
-
+    n = length(keys(D))
     # invoke core algorithm
     _dbscan(D, convert(T, eps), minpts, 1:n)
 end
--- a/src/dbscan.jl
+++ b/src/dbscan.jl
@@ -6,7 +6,7 @@
 #       A density-based algorithm for discovering clusters
 #       in large spatial databases with noise. 1996.
 #
-
+import Arrow
 """
     DbscanResult <: ClusteringResult
 
--- a/src/dbscan.jl
+++ b/src/dbscan.jl
@@ -66,6 +66,15 @@ function dbscan(D::AbstractMatrix{T}, eps::Real, minpts::Int) where T<:Real
     _dbscan(D, convert(T, eps), minpts, 1:n)
 end
 
+function dbscan(D::Arrow.Table, ϵ::Real, minpts::Int; T = Base.Float64)
+    # check arguments    
+    eps > 0 || throw(ArgumentError("ϵ must be a positive value ($eps given)."))
+    minpts >= 1 || throw(ArgumentError("minpts must be positive integer ($minpts given)."))
+
+    # invoke core algorithm
+    _dbscan(D, convert(T, eps), minpts, 1:n)
+end
+
 function _dbscan(D::AbstractMatrix{T}, eps::T, minpts::Int, visitseq::AbstractVector{Int}) where T<:Real
     n = size(D, 1)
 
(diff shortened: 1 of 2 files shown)


alyst:
Thanks for the PR! Supporting large datasets is definitely very interesting.

We just have to be conservative with introducing additional dependencies to *Clustering.jl*.
From the first glance, it doesn't look like you are using anything *Arrow.jl*-specific.
Is it possible to remove *Arrow.jl* as a direct dependency and replace it with *Tables.jl* or *TableTraits.jl*?
You can still add *Arrow.jl* as a test dependency and add the tests operate with Arrow format files.

src/dbscan.jl
```diff
@@ -94,6 +103,36 @@ function _dbscan(D::AbstractMatrix{T}, eps::T, minpts::Int, visitseq::AbstractVe
     return DbscanResult(seeds, assignments, counts)
 end
 
+#Arrow.Table support
+function _dbscan(D::Arrow.Table, eps::T, minpts::Int, visitseq::AbstractVector{Int}) where T<:Real
```
alyst, line 107: I don't see any differences between this `_dbscan()` overload and the generic `_dbscan()`.
You don't have to duplicate the function to support another input type, you can just relax the type constraints of the existing one (e.g. `D::Union{AbstractMatrix, Arrow.Table}`).
The only difference between `AbstractMatrix` and `Arrow.Table` that I see is the column access: `view(D, :, p)` for the `AbstractMatrix`, and `D[p]` for the *Arrow.Table*.
Is it just this or there is something else?
Have you checked if there is a way to support `view(D, :, p)` for `Arrow.Table`?
Would `view(transpose(D), :, p)` or something like that work?

ma-seefelder:
Thanks for your comments on my PR. Unfortunately, it is not possible in my opinion to just relax the type constraints of the existing function. There are several changes to the functions dbscan, _dbscan, and _dbs_region_query, (lines 73, 108, 152). For the function _dbs_expand_cluster!, I now just relaxed the type constraints as you suggested. If preferred, I can implement for the other functions versions with relaced type constraints and if statements like:

```
function _dbs_region_query(D::Union{AbstractMatrix{T}, Arrow.Table}, p::Int, eps::T) where T<:Real
   if typeof(D) == Arrow.Table
       n = length(keys(D))
       dists = D[p]
   else
       n = size(D,1)
       dists = view(D,:,p)
   end

  nbs = Int[]
   
 for i = 1:n
        @inbounds if dists[i] < eps
            push!(nbs, i)
        end
    end
    return nbs::Vector{Int}
end

```

All the changes are because the views function is not implemented for Arrow.Tables without prior conversion of the Arrow.Table to a DataFrame or that the size(D::Arrow.Table) is also not defined. Conversion to a DataFrame or Tables.jl object copies the whole data and tries to store it in memory.

alyst:
> All the changes are because the views function is not implemented for Arrow.Tables

Still, if the only difference is the row access,  you can define e.g.`_sample(D::AbstractMatrix, ix::Integer) = view(D, :, ix)`, and `_sample(D::Tables.AbstractColumns, ix::Integer) = D[ix]` and use the `_sample()` function from the single implementation of the algorithm. The same goes for the utility function to get the number of samples. Note that in my example I use the base type from *Tables.jl*, of which `Arrow.Table` is a subtype. That way we don't need to require *Arrow.jl* dependency.
Also, quite often the explicit test for the variable type inside the function body indicates suboptimal design.

alyst:
One problem that I see with `Arrow.Table` support after thinking a bit more is that it breaks the convention that samples are stored column-wise, because for the `Arrow.Table` they are stored row-wise.

An alternative approach that would not require modifying the DBSCAN code and will keep the column-wise sample convention would be to define the `Arrow.Table` wrapper that transposes the rows:
```julia
struct TransposedColumns{T, C <: Tables.AbstractColumns} <: AbstractMatrix{T}
   columns::C
end

Base.transpose(cols::Tables.AbstractColumns) = TransposedColumns{eltype(cols)}(cols)
Base.view(tcols::TransposedColumns, ::Colon, ix::Any) = tcols.columns[ix]
Base.size(tcols::TransposedColumns) = ...
....
```

So you can just `dbscan(transpose(table))`.

Issue #221, opened 2021-08-31
DBSCAN docs - consistency between 2 APIs


BioTurboNick:
E.g. How does minpts in the distance matrix version map to the kwargs in the adjacency list version?

alyst:
*dbscan()* API has been refractored

Issue #200, opened 2020-07-23
DBscan Adjacency Lists have repeated clusters


MRIO:
Dear all,

I have been attempting to using dbscan on a 2D array (size: 2 x 1105 ) of point positions (results of UMAP), but i get this strange result, where there are repeated DbscanClusters inside the output array:
```

13-element Array{DbscanCluster,1}:
 DbscanCluster(17, [4, 12, 84, 90, 94, 675, 676, 737, 873, 965], [27, 108, 177, 880, 954, 1050, 1067])
 DbscanCluster(10, Int64[], [46, 48, 51, 57, 188, 225, 226, 228, 270, 542])
 DbscanCluster(11, [48, 51, 228], [46, 49, 57, 188, 225, 226, 270, 542])
 DbscanCluster(14, [418, 759, 832, 988, 1046], [830, 831, 855, 865, 989, 991, 996, 1021, 1070])
 DbscanCluster(10, Int64[], [624, 654, 664, 803, 805, 821, 859, 987, 1057, 1069])
 DbscanCluster(10, Int64[], [624, 654, 664, 803, 805, 821, 859, 987, 1057, 1069])
 DbscanCluster(10, Int64[], [624, 654, 664, 803, 805, 821, 859, 987, 1057, 1069])
 DbscanCluster(10, Int64[], [624, 654, 664, 803, 805, 821, 859, 987, 1057, 1069])
 DbscanCluster(10, Int64[], [624, 654, 664, 803, 805, 821, 859, 987, 1057, 1069])
 DbscanCluster(10, Int64[], [624, 654, 664, 803, 805, 821, 859, 987, 1057, 1069])
 DbscanCluster(10, Int64[], [624, 654, 664, 803, 805, 821, 859, 987, 1057, 1103])
 DbscanCluster(10, Int64[], [624, 654, 664, 803, 805, 821, 859, 987, 1057, 1103])
 DbscanCluster(11, [1057], [624, 654, 664, 803, 805, 821, 859, 987, 1069, 1103])
```
The data on which I'm applying dbscan looks sane (and there are no repeats):

<img width="570" alt="image" src="https://user-images.githubusercontent.com/8012564/88263353-939d9d00-ccd2-11ea-9a6d-9668fdf3251d.png">

Any idea why this is happening?

alyst:
There was a bug in coordinate-based `dbscan()` implementation: it created a cluster even if there are no seeds/core points.
[...]

PR #248, opened 2023-03-20
Refactor dbscan()

alyst:
We have two implementations of dbscan(): one is distance matrix-based, another is NNTree-based. As noted in #221, they have inconsistent APIs and return different type.
This PR consolidates the two implementations (based on NNTree one), fixes some behaviour differences and sets deprecations:
  * `dbscan(dist, eps, minpts)` is deprecated in favor of `dbscan(dist, radius, metric=nothihg, min_neighbors=minpts)`
  * ~~`dbscan()` adds `return_type=` parameter, which is `Vector{DbscanCluster}` for `dbscan(points, ...)`, but this behaviour is deprecated in favor of `return_type=DbscanResult`.~~
    `dbscan()` now always returns `DbscanResult`.  This is a breaking change for NN-tree-based implementation, which returned `Vector{DbscanCluster}`.
 * fixes a bug in NNTree-based implementation -- the check that the cluster core is not empty was missing. See e.g. #200
 * *radius*=*eps*=0 is allowed for both implementations
 * *min_neighbors* now includes the point itself (as in the original pseudocode and distance matrix-based impl)
  * *metric* allows specifying non-Euclidean metrics; when `metric=nothing`, it signifies that the matrix provided is the distance matrix.


alyst:
@ararslan Sorry for bothering again. I've tried two fix the inconsistencies of the two `dbscan()` implementations we have: they differ in return type and positional arguments. It required creating two deprecations to maintain backward compatibility. But potentially there are other alternatives for the update API and deprecation path. Let me know what you think.

alyst:
> I feel like the latter might actually be easiest for both users and maintainers. Thoughts?

Thanks for allowing me the breaking change :)
It is a much more straightforward path + NN-tree based *dbscan()* (the one that returned *Vector{DbscanCluster}*) gave incorrect results anyway,
so the breaking change will alert the users that they should expect very different clustering.

Issue #253, opened 2023-06-06
Providing additional intrinsic evaluation metrices


jaksle:
This is more of a question than an issue. I wanted to time series clustering using Clustering.jl and I needed intrinsic evaluation measures in order to determine optimal number of clusters. Clustering.jl offers only silhouettes in this regard so I implemented myself popular ones: Calinski-Harabasz, Davies-Boulden, Xie-Beni indices. Would it be useful if I add them to Clustering.jl?

The formulas are easy so it would not require a lot of work, it's just that first I would like to be sure there is an interest in that.

nomadbl:
Not a maintainer here or anything, but sounds interesting to me :smile:

alyst:
Sure, this sounds very useful! One constraint, though, is that we don't want *Clustering.jl* to be a super-package for anything related to unsupervised learning as that would complicate maintenance and unnecessary bloat user projects.
So new features that would require extra dependencies, esp. outside of *github.com/JuliaXXX* ecosystem, or features that add a lot of new relatively complex code, should rather be implemented as separate packages.

If these new metrics fit the above criteria, the PR is welcome! :)

jaksle:
No new dependencies needed, no complex code, these quality indices have definitions simpler than silhouettes.  Bloat is not my goal, but the few most popular indices for hard and soft clustering I hope would come in handy.

So great, now I am busy but next month I will try check and adjust my code :)

alyst:
Implemented in #257

PR #255, opened 2023-06-25
Distributed silhouette

nomadbl:
This PR implements a distributed implementation of silhouette score
https://arxiv.org/pdf/2303.14102.pdf

This implementation only supports Square Euclidean metric at the moment, but the extension to some (not all) other metrics is straightforward. See the paper.

This PR includes:
- Docs incl. usage example (see below)
- Implementation
- a bit of a refactor of the existing silhouette code (break it down into more functions to reuse in the new distributed implementation).
- Unit tests of the new implementation, including a comparison with the usual implementation.

Example including benchmarking
```
Julia> k=10; d=3; bs=100; n=10000;
Julia> pre = SilhouettesDistsPrecompute(k, d, Float32);
Julia> x = reshape(collect(Float32, 1:30000), 3, n); # direct computation is impractical on such a big dataset
Julia> a = reshape(repeat(collect(1:k),trunc(Int, n/k)), n);
Julia> batches_x = eachslice(reshape(x, 3, trunc(Int, n/bs), bs); dims=3); batches_a = eachslice(reshape(a, trunc(Int, n/bs), bs); dims=2);
Julia> @time [silhouettes_precompute_batch!(k, aa, xx, pre) for (xx, aa) in zip(batches_x, batches_a)]; # precompute vectors on big data
0.838358 seconds (2.06 M allocations: 141.495 MiB, 4.24% gc time, 99.00% compilation time)
Julia> @time sil = vcat([silhouettes(xx, aa, pre) for (xx, aa) in zip(batches_x, batches_a)]); # calculate silhouette scores in batched fashion
0.049759 seconds (53.78 k allocations: 4.440 MiB, 98.54% compilation time)
Julia> size(sil)
(10000,)
```

### DIFF
--- a/src/Clustering.jl
+++ b/src/Clustering.jl
@@ -83,6 +83,7 @@ module Clustering
     include("fuzzycmeans.jl")
 
     include("counts.jl")
+    include("cluster_distances.jl")
     include("silhouette.jl")
     include("randindex.jl")
     include("varinfo.jl")
--- a/src/Clustering.jl
+++ b/src/Clustering.jl
@@ -47,7 +47,7 @@ module Clustering
     counts, # reexport StatsBase.counts
 
     # silhouette
-    silhouettes,
+    silhouettes, SilhouettesDistsPrecompute, silhouettes_precompute_batch!,
 
     # varinfo
     varinfo,
--- a/src/Clustering.jl
+++ b/src/Clustering.jl
@@ -47,7 +47,7 @@ module Clustering
     counts, # reexport StatsBase.counts
 
     # silhouette
-    silhouettes, SilhouettesDistsPrecompute, silhouettes_precompute_batch!,
+    silhouettes, SqEuclideanPrecomputedSilhouettes, silhouettes_precompute_batch!,
 
     # varinfo
     varinfo,
--- a/src/Clustering.jl
+++ b/src/Clustering.jl
@@ -47,7 +47,7 @@ module Clustering
     counts, # reexport StatsBase.counts
 
     # silhouette
-    silhouettes, SqEuclideanPrecomputedSilhouettes, silhouettes_precompute_batch!,
+    silhouettes, SqEuclideanPrecomputedSilhouettes, precompute_silhouettes,
 
     # varinfo
     varinfo,
--- a/src/Clustering.jl
+++ b/src/Clustering.jl
@@ -47,7 +47,7 @@ module Clustering
     counts, # reexport StatsBase.counts
 
     # silhouette
-    silhouettes, PrecomputeSilhouettes, silhouettes_cache,
+    silhouettes,
 
     # varinfo
     varinfo,
--- a/src/Clustering.jl
+++ b/src/Clustering.jl
@@ -47,7 +47,7 @@ module Clustering
     counts, # reexport StatsBase.counts
 
     # silhouette
-    silhouettes, SqEuclideanPrecomputedSilhouettes, precompute_silhouettes,
+    silhouettes, SqEuclideanPrecomputedSilhouettes, silhouettes_cache,
 
     # varinfo
     varinfo,
--- a/src/Clustering.jl
+++ b/src/Clustering.jl
@@ -47,7 +47,7 @@ module Clustering
     counts, # reexport StatsBase.counts
 
     # silhouette
-    silhouettes, SqEuclideanPrecomputedSilhouettes, silhouettes_cache,
+    silhouettes, PrecomputeSilhouettes, silhouettes_cache,
 
     # varinfo
     varinfo,
--- a/src/cluster_distances.jl
+++ b/src/cluster_distances.jl
@@ -1,0 +1,226 @@
+"""
+Base type for efficient computation of average(mean) distances
+from the cluster centers to a given point.
+
+The descendant types should implement the following methods:
+  * `update!(dists, assignments, points)`: update the internal
+    state of `dists` with point coordinates and their assignments to the clusters
+  * `sumdistances(dists, points, indices)`: compute the sum of
+    distances from all `dists` clusters to `points`
+"""
+abstract type ClusterDistances{T} end
+
+# create empty ClusterDistances object for a given metric
+# and update it with a given clustering
+# if batch_size is specified, the updates are done in point batches of given size
+function ClusterDistances(metric::SemiMetric,
+                 assignments::AbstractVector{<:Integer},
+                 points::AbstractMatrix{<:Real},
+                 batch_size::Union{Integer, Nothing} = nothing)
+    update!(ClusterDistances(eltype(points), metric, length(assignments), size(points, 1),
+                             maximum(assignments)),
+            assignments, points, batch_size)
+end
+
+ClusterDistances(metric, R::ClusteringResult, args...) =
+    ClusterDistances(metric, assignments(R), args...)
+
+# fallback implementations of ClusteringDistances methods
+
+cluster_sizes(dists::ClusterDistances) = dists.cluster_sizes
+nclusters(dists::ClusterDistances) = length(cluster_sizes(dists))
+
+update!(dists::ClusterDistances,
+        assignments::AbstractVector, points::AbstractMatrix) =
+    error("update!(dists::$(typeof(dists))) not implemented")
+
+sumdistances(dists::ClusterDistances,
+             points::Union{AbstractMatrix, Nothing},
+             indices::Union{AbstractVector{<:Integer}, Nothing}) =
+    error("sumdistances(dists::$(typeof(dists))) not implemented")
+
+# average distances from each cluster to each point, nclusters×n matrix
+function meandistances(dists::ClusterDistances,
+                       assignments::AbstractVector{<:Integer},
+                       points::Union{AbstractMatrix, Nothing},
+                       indices::AbstractVector{<:Integer})
+    @assert length(assignments) == length(indices)
+    (points === nothing) || @assert(size(points, 2) == length(indices))
+    clu_to_pt = sumdistances(dists, points, indices)
+    clu_sizes = cluster_sizes(dists)
+    @assert length(assignments) == length(indices)
+    @assert size(clu_to_pt) == (length(clu_sizes), length(assignments))
+
+    # normalize distances by cluster sizes
+    @inbounds for j in eachindex(assignments)
+        for (i, c) in enumerate(clu_sizes)
+            if i == assignments[j]
+                c -= 1
+            end
+            if c == 0
+                clu_to_pt[i,j] = 0
+            else
+                clu_to_pt[i,j] /= c
+            end
+        end
+    end
+    return clu_to_pt
+end
+
+# wrapper for ClusteringResult
+update!(dists::ClusterDistances, R::ClusteringResult, args...) =
+    update!(dists, assignments(R), args...)
+
+# batch-update silhouette dists (splitting the points into chunks of batch_size size)
+function update!(dists::ClusterDistances,
+                 assignments::AbstractVector{<:Integer}, points::AbstractMatrix{<:Real},
+                 batch_size::Union{Integer, Nothing})
+    n = size(points, 2)
+    ((batch_size === nothing) || (n <= batch_size)) &&
+        return update!(dists, assignments, points)
+
+    for batch_start in 1:batch_size:n
+        batch_ixs = batch_start:min(batch_start + batch_size - 1, n)
+        update!(dists, view(assignments, batch_ixs), view(points, :, batch_ixs))
+    end
+    return dists
+end
+
+# generic ClusterDistances implementation for an arbitrary metric M
+# if M is Nothing, point_dists is an arbitrary matrix of point distances
+struct SimpleClusterDistances{M, T} <: ClusterDistances{T}
+    metric::M
+    cluster_sizes::Vector{Int}
+    assignments::Vector{Int}
+    point_dists::Matrix{T}
+
+    SimpleClusterDistances(::Type{T}, metric::M,
+                           npoints::Integer, nclusters::Integer) where {M<:Union{SemiMetric, Nothing}, T<:Real} =
+        new{M, T}(metric, zeros(Int, nclusters), Vector{Int}(),
+                  Matrix{T}(undef, npoints, npoints))
+
+    # reuse given points matrix
+    function SimpleClusterDistances(
+        metric::Nothing,
+        assignments::AbstractVector{<:Integer},
+        point_dists::AbstractMatrix{T}
+    ) where T<:Real
+        n = length(assignments)
+        size(point_dists) == (n, n) || throw(DimensionMismatch("assignments length ($n) does not match distances matrix size ($(size(point_dists)))"))
+        issymmetric(point_dists) || throw(ArgumentError("point distances matrix must be symmetric"))
+        clu_sizes = zeros(Int, maximum(assignments))
+        @inbounds for cluster in assignments
+            clu_sizes[cluster] += 1
+        end
+        new{Nothing, T}(metric, clu_sizes, assignments, point_dists)
+    end
+end
+
+# fallback ClusterDistances constructor
+ClusterDistances(::Type{T}, metric::Union{SemiMetric, Nothing},
+                 npoints::Union{Integer, Nothing}, dims::Integer, nclusters::Integer) where T<:Real =
+    SimpleClusterDistances(T, metric, npoints, nclusters)
+
+# when metric is nothing, points is the matrix of distances
+function ClusterDistances(metric::Nothing,
+                          assignments::AbstractVector{<:Integer},
+                          points::AbstractMatrix,
+                          batch_size::Union{Integer, Nothing} = nothing)
+    (batch_size === nothing) || (size(points, 2) > batch_size) ||
+        error("batch-updates of distance matrix-based ClusterDistances not supported")
+    SimpleClusterDistances(metric, assignments, points)
+end
+
+function update!(dists::SimpleClusterDistances{M},
+                 assignments::AbstractVector{<:Integer},
+                 points::AbstractMatrix{<:Real}) where M
+    @assert length(assignments) == size(points, 2)
+    check_assignments(assignments, nclusters(dists))
+    append!(dists.assignments, assignments)
+    n = size(dists.point_dists, 1)
+    length(dists.assignments) == n ||
+        error("$(typeof(dists)) does not support batch updates: $(length(assignments)) points given, $n expected")
+    @inbounds for cluster in assignments
+        dists.cluster_sizes[cluster] += 1
+    end
+
+    if M === Nothing
+        size(point_dists) == (n, n) ||
+            throw(DimensionMismatch("points should be a point-to-point distances matrix of ($n, $n) size, $(size(points)) given"))
+        copy!(dists.point_dists, point_dists)
+    else
+        # metric-based SimpleClusterDistances does not support batched updates
+        size(points, 2) == n ||
+            throw(DimensionMismatch("points should be a point coordinates matrix with $n columns, $(size(points, 2)) found"))
+        pairwise!(dists.metric, dists.point_dists, points, dims=2)
+    end
+
+    return dists
+end
+
+# this function returns matrix r nclusters x n, such that
+# r[i, j] is the sum of distances from all i-th cluster points to the point indices[j]
+function sumdistances(dists::SimpleClusterDistances,
+                      points::Union{AbstractMatrix, Nothing}, # unused as distances are already in point_dists
+                      indices::AbstractVector{<:Integer})
+    T = eltype(dists.point_dists)
+    n = length(dists.assignments)
+    S = typeof((one(T)+one(T))/2)
+    r = zeros(S, nclusters(dists), n)
+    @inbounds for (jj, j) in enumerate(indices)
+        for i = 1:j-1
+            r[dists.assignments[i], jj] += dists.point_dists[i,j]
+        end
+        for i = j+1:n
+            r[dists.assignments[i], jj] += dists.point_dists[i,j]
+        end
+    end
+    return r
+end
+
+# uses the method from "Distributed Silhouette Algorithm: Evaluating Clustering on Big Data"
+# https://arxiv.org/abs/2303.14102
+# for SqEuclidean point distances
+struct SqEuclideanClusterDistances{T} <: ClusterDistances{T}
+    cluster_sizes::Vector{Int} # [nclusters]
+    Y::Matrix{T} # [dims, nclusters], the first moments of each cluster (sum of point coords)
+    Ψ::Vector{T} # [nclusters], the second moments of each cluster (sum of point coord squares)
+
+    SqEuclideanClusterDistances(::Type{T}, npoints::Union{Integer, Nothing}, dims::Integer,
+                                nclusters::Integer) where T<:Real =
+        new{T}(zeros(Int, nclusters), zeros(T, dims, nclusters), zeros(T, nclusters))
+end
+
+ClusterDistances(::Type{T}, metric::SqEuclidean, npoints::Union{Integer, Nothing},
+                 dims::Integer, nclusters::Integer) where T<:Real =
+    SqEuclideanClusterDistances(T, npoints, dims, nclusters)
+
+function update!(dists::SqEuclideanClusterDistances,
+                 assignments::AbstractVector{<:Integer},
+                 points::AbstractMatrix{<:Real})
+    # x dims are [D,N]
+    d, n = size(points)
+    k = length(cluster_sizes(dists))
+    check_assignments(assignments, k)
+    n == length(assignments) || throw(DimensionMismatch("points count ($n) does not match assignments length $(length(assignments)))"))
+    d == size(dists.Y, 1) || throw(DimensionMismatch("points dims ($(size(points, 1))) do no must match ClusterDistances dims ($(size(dists.Y, 1)))"))
+    # precompute moments and update counts
+    @inbounds for (i, cluster) in enumerate(assignments)
+        point = view(points, :, i) # switch to eachslice() once Julia-1.0 support is dropped
+        dists.cluster_sizes[cluster] += 1
+        dists.Y[:, cluster] .+= point
+        dists.Ψ[cluster] += sum(abs2, point)
+    end
+    return dists
+end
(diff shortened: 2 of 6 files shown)


src/silhouette.jl
```diff
@@ -18,48 +84,11 @@ function sil_aggregate_dists(k::Int, a::AbstractVector{Int}, dists::AbstractMatr
...
-    k >= 2 || throw(ArgumentError("silhouettes() not defined for the degenerated clustering with a single cluster."))
-    for j = 1:n
-        (1 <= assignments[j] <= k) || throw(ArgumentError("Bad assignments[$j]=$(assignments[j]): should be in 1:$k range."))
-    end
-    sum(counts) == n || throw(ArgumentError("Mismatch between assignments ($n) and counts (sum(counts)=$(sum(counts)))."))
-    size(dists) == (n, n) || throw(DimensionMismatch("The size of a distance matrix ($(size(dists))) doesn't match the length of assignment vector ($n)."))
-
-    # compute average distance from each cluster to each point --> r
-    r = sil_aggregate_dists(k, assignments, dists)
+function normalize_aggregate_dists!(r, assignments, counts, k, n)
```
alyst, line 87: I suggest avoiding abbreviated function names as it makes the code a bit more readable.
```suggestion
function normalize_aggregated_distances!(r::AbstractMatrix, assignments::AbstractVector, counts::AbstractVector)
    ....
end
```
we don't need to pass `k` and `n`, they are just the vector lengths.

src/silhouette.jl
```diff
@@ -70,13 +99,48 @@ function silhouettes(assignments::AbstractVector{<:Integer},
...
+
+function sil_aggregate_dists_normalized(assignments::AbstractVector{<:Integer},
+                                        counts::AbstractVector{<:Integer},
+                                        dists::AbstractMatrix{T}) where T<:Real
+    n = length(assignments)
+    k = length(counts)
+    k >= 2 || throw(ArgumentError("silhouettes() not defined for the degenerated clustering with a single cluster."))
+    for j = 1:n
+        (1 <= assignments[j] <= k) || throw(ArgumentError("Bad assignments[$j]=$(assignments[j]): should be in 1:$k range."))
+    end
```
alyst, line 113:
```suggestion
    k >= 2 || throw(ArgumentError("silhouettes() not defined for the degenerated clustering with a single cluster."))
    check_assignments(assignments, k)
```

src/silhouette.jl
```diff
@@ -70,13 +99,48 @@ function silhouettes(assignments::AbstractVector{<:Integer},
...
+    size(x, 1) == size(pre.Y, 1) || throw(ArgumentError("input features dimension does not match with precomputation on dimension 1"))
+
+    # compute average distance from each cluster to each point --> r
+    ξx = sum(x.^2; dims=1) # [1,n]
+    r = pre.counts .* ξx .+ pre.Ψ .- 2 .* transpose(pre.Y) * x # r is [k, n]
+    r = normalize_aggregate_dists!(r, assignments, pre.counts, k, n)
+    return r
+end
+
+function silhouettes_given_dist(r::Matrix{T}, assignments::AbstractVector{Int}, counts::Vector{Int}) where {T}
```
alyst, line 136:
```suggestion
# compute silhouette scores for each point given a matrix of cluster-to-point distances
function silhouette_scores(cluster_to_point::AbstractMatrix{<:Real}, assignments::AbstractVector{<:Integer}, counts::AbstractVector{<:Integer})
```

src/silhouette.jl
```diff
@@ -70,13 +99,48 @@ function silhouettes(assignments::AbstractVector{<:Integer},
             end
         end
     end
+    return r
+end
+
+function sil_aggregate_dists_normalized(assignments::AbstractVector{<:Integer},
```
alyst, line 105: I'm not sure we need these function in addition to `sil_aggregate_dist`, the only thing it does is calling `normalize...()`

alyst (changes requested):
Thanks for the PR!
There are two major considerations:
  * could you please provide a benchmark similar to the original paper that shows how faster is this method in comparison with the original implementation?
  * since @jaksle suggested to contribute alternative clustering evaluation metrics (#253), could you please coordinate with him that this PR would not clash with his code? I've provided some initial code review, but to avoid doing extra work, I suggest figuring out this question first.

jaksle:
Interesting algorithm! There should be no conflicts with what I plan to implement, other metrics do not need to use silhouettes code in any way.

src/silhouette.jl
```diff
@@ -18,48 +84,11 @@ function sil_aggregate_dists(k::Int, a::AbstractVector{Int}, dists::AbstractMatr
...
-    size(dists) == (n, n) || throw(DimensionMismatch("The size of a distance matrix ($(size(dists))) doesn't match the length of assignment vector ($n)."))
-
-    # compute average distance from each cluster to each point --> r
-    r = sil_aggregate_dists(k, assignments, dists)
+function normalize_aggregate_dists!(r, assignments, counts, k, n)
     # from sum to average
     @inbounds for j = 1:n
         for i = 1:k
-            c = counts[i]
+            c = copy(counts[i])
```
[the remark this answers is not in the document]
nomadbl: I did this because the variable c is modified during the loop, which I was afraid might modify the counts array in the precomputed variables struct.
I can test without this to see the results are consistent.

src/silhouette.jl
nomadbl: The purpose here was to test that the point to cluster distances were consistent with the baseline code, before the normalization.
Since this is not something the user needs to worry about, I think we should leave this, as this provides another important layer of testing.

nomadbl:
Thanks for the thorough review!
I'll make the changes as you suggested and provide details in the comments as necessary.

src/silhouette.jl
```diff
@@ -1,5 +1,71 @@
 # Silhouette
 
+mutable struct SilhouettesDistsPrecompute{T}
+    k::Int
+    d::Int
+    counts::Matrix{Int} #[k, 1]
+    Ψ::Matrix{T} #[k, 1]
+    Y::Matrix{T} #[d, k]
+end
```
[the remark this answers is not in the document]
nomadbl: 1. :heavy_check_mark: 
2. :heavy_check_mark: 
3. What do you think of `cluster_first_moment` and `cluster_second_moment`? I think they are more descriptive here.
4. See the example in the docstring of `silhouettes_precompute_batch!`. As you can see, the user needs to hold an instance of `SqEuclideanPrecomputedSilhouettes` as he runs through the data set. The data points fed to `silhouettes` are in general a tiny fraction of the dataset, so he needs to keep a reference to the precomputed values as he goes.
5. The purpose of the matrix vs vector was for the broadcasting operations, so that there is no ambiguity there. What of the case of i.e. `k == d`...

src/silhouette.jl
```diff
@@ -1,5 +1,71 @@
...
+
+"""
+SilhouettesDistsPrecompute(k::Int, d::Int, ::Type{T})
+Precomputations container for [`silhouettes_precompute_batch!`](@ref).
+See also [`silhouettes`](@ref), [`silhouettes_precompute_batch!`](@ref)
+"""
+SilhouettesDistsPrecompute(k::Int, d::Int, ::Type{T}) where T<:Real= SilhouettesDistsPrecompute{T}(k, d, 
+                                                                                      zeros(Int, k, 1),
+                                                                                      zeros(T, k, 1), 
+                                                                                      zeros(T, d, k))
```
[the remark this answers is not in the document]
nomadbl: Internal vs external: I thought internal constructors were meant to impose restrictions for the struct instances? This constructor is just meant as a useful shortcut.

See the comment on exporting in the above conversation.

src/silhouette.jl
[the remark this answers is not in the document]
alyst: `Int` is a primitive type, the assignment already copies the bits, not the reference to the array element.

src/silhouette.jl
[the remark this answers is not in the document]
alyst: Exactly, this constructor imposes the restriction on k, n, the dimensions of the matrices and their initial content, so it makes to make it internal.

src/silhouette.jl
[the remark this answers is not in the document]
alyst: 3. I'm not sure I understand the code well enough to suggest the proper names. :) Too long names names also degrade code readability. Single letter vars are OK if there is a comment with the description, but for fields like k and d the names nclusters and dims would be better IMO.
4. Sure, you can have an exported method that returns this object, but there is no need to export the type of object, as the user does not need to access its fields. If you want, you may want to declare an abstract *AbstractPrecomputedSilhouettes{T}* type export it, and inherit all concrete types from it. But at the moment if it's an overkill as we have a single metric, and I suspect uses of "streaming" API wiggle be rare.
5. Sure, for broadcasting you might be required to reshape it, before using, that's what I've meant. But reshaping just creates a temporary array that references the same memory, so it's cheap. For any other operations keeping the dimensions as lean as possible is the most practical solution.

src/silhouette.jl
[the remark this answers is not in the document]
alyst: You only need to copy if the elements are of mutable type (mutable struct, arrays), or even deepcopy() if the fields are mutable, because the assignment just assigns the reference to the same memory.

src/silhouette.jl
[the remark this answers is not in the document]
nomadbl: These points should be resolved, but please take a look and resolve this conversation if all is well.

src/silhouette.jl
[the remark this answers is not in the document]
nomadbl: These points should be resolved, but please take a look and resolve this conversation if all is well.

nomadbl:
I updated the API in the spirit of @alyst 's suggestion.
There are mainly two additions:
- Silhouettes uses the streaming implementation on supported metrics by default (can be overriden, and falls back to basic implementation for unsupported metrics)
- The precomputation struct can be used to call the precomputation method as a functor. This simplifies things for the user and allows easy extension later, because:
- A new user facing function `silhouettes_cache(::Type, metric::Union{Metric, SemiMetric}, nclusters, dims)` handles the constructors so the user specifies the metric he is using and gets the correct struct (he is also thrown an error in case of unsupported metric).
- etc, see changes in the code.

Also, I added a benchmark comparing the vanilla and streaming implementation. You can see the results on my machine below, clearly displaying O(N) vs O(N^2) behavior. Note the distances calculation is included in the benchmark time, I wasn't sure if this was appropriate but left it there for now.

```
"silhouette" => 4-element BenchmarkTools.BenchmarkGroup:
	  tags: []
	  "n=20,000" => 2-element BenchmarkTools.BenchmarkGroup:
		  tags: []
		  "with_precalculation" => Trial(2.102 ms)
		  "without_precalculation" => Trial(4.916 s)
	  "n=10,000" => 2-element BenchmarkTools.BenchmarkGroup:
		  tags: []
		  "with_precalculation" => Trial(1.027 ms)
		  "without_precalculation" => Trial(913.840 ms)
	  "n=1,000" => 2-element BenchmarkTools.BenchmarkGroup:
		  tags: []
		  "with_precalculation" => Trial(91.534 μs)
		  "without_precalculation" => Trial(3.986 ms)
	  "n=100" => 2-element BenchmarkTools.BenchmarkGroup:
		  tags: []
		  "with_precalculation" => Trial(10.111 μs)
		  "without_precalculation" => Trial(42.006 μs)
```

src/silhouette.jl
```diff
@@ -101,12 +171,60 @@ function silhouettes(assignments::AbstractVector{<:Integer},
...
+
+# References
+> Peter J. Rousseeuw (1987). *Silhouettes: a Graphical Aid to the
+> Interpretation and Validation of Cluster Analysis*. Computational and
+> Applied Mathematics. 20: 53–65.
+> Marco Gaido (2023). Distributed Silhouette Algorithm: Evaluating Clustering on Big Data 
+"""
+function silhouettes(assignments::AbstractVector{<:Integer},
+                     counts::AbstractVector{<:Integer},
+                     dists::AbstractMatrix{T}) where T<:Real
```
alyst, line 217:
```suggestion
function silhouettes(assignments::AbstractVector{<:Integer}, points::AbstractMatrix;
                     metric::Union{Metric, SemiMetric, Nothing} = nothing, 
                     method::Symbol = :default)
```

This is how I see the updated `silhouettes()` API. It is analogous to `dbscan()`, so you can take a look at the logic there. 
Basically, if `metric = nothing` (the default), `points` are treated as distances, otherwise they are treated as coordinates, and `metric` defines how distances are calculated.
`method` defines what implementation to use, the options are:
  * `:classic`: the original implementation
  * `:cached` (alternatively, `:distributed`; you may have better variants): your implementation. If this method is not compatible with the `metric`, the exception should be thrown. (Not-supported metric would throw automatically).
  * `:default`: the best method is selected, which would be `:cached` for supported metrics, `:classic` otherwise

This is the main method, the aliases that expect `ClusteringResult` should pass through the keywords.

src/silhouette.jl
```diff
@@ -101,12 +171,60 @@ function silhouettes(assignments::AbstractVector{<:Integer},
...
+
+    cluster_to_point = sil_aggregate_distances_normalized(assignments, counts, dists)
+    return silhouette_scores(cluster_to_point, assignments, counts)
+end
+
+function silhouettes(x::AbstractMatrix{T}, assignments::AbstractVector{<:Integer}, pre::PrecomputeSilhouettes) where T<:Real
```
alyst, line 223:
```suggestion
function silhouettes(pre::PrecomputeSilhouettes, assignments::AbstractVector{<:Integer}, points::AbstractMatrix{T}) where T<:Real
```
If it actually distances or point coordinates? In the former case the parameter should be called `dists`.

src/utils.jl
```diff
@@ -76,3 +76,10 @@ function updatemin!(r::AbstractArray, x::AbstractArray)
     end
     return r
 end
+
+function check_assignments(assignments, nclusters)
```
alyst, line 80:
```suggestion
function check_assignments(assignments::AbstractVector{<:Integer}, nclusters::Integer)
```

src/utils.jl
```diff
@@ -76,3 +76,10 @@ function updatemin!(r::AbstractArray, x::AbstractArray)
...
+    nclusters >= 2 || throw(ArgumentError("silhouettes() not defined for the degenerated clustering with a single cluster."))
```
alyst, line 81: Ah, this line is `silhouettes()`-specific, it should be moved out of generic checking function to `silhouettes()` code.

alyst (changes requested):
Thanks for adding the benchmarking code!

Could you please share the plots (preferred) or textual output that compares the current `silhouettes()` performance with the new "all-at-once"? We have to confirm that there is an improvement, this is a prerequisite for merging the PR.

I see that you have advanced with the streaming/precomputed/distributed implementation. I understand your motivation, but let's move it out of scope of this PR:
  * *Clustering.jl* provides basic unsupervised learning algorithms, none of which were specifically optimized for very large datasets or distributed processing. Providing and maintaining such capabilities would require much more resources than the community can dedicate. Also, if the users cannot cluster their very large datasets with *Clustering.jl*, it would be of little use if *silhouettes()* can evaluate the quality of these non-existing clusters.
  * To really exploit the benefits of "streaming" silhouettes, you would need to support multi-processing or multi-threading, which would complicate the code even more. In fact, under-the-hood, this PR would provide `precompute_silhouettes()!` and `silhouettes(precomputed::PrecomputedSilhouettes)` functions, so you can still explore batched computation, multithreading and multiprocessing on the side.

So I suggest this PR only uses "all-at-once" for `silhouettes()` (given it is faster than the classical one).

Let's try to iron out the API for the next iteration. Afterwards it should be mostly cosmetic things.

src/Clustering.jl
```diff
@@ -47,7 +47,7 @@ module Clustering
     counts, # reexport StatsBase.counts
 
     # silhouette
-    silhouettes,
+    silhouettes, PrecomputeSilhouettes, silhouettes_cache,
```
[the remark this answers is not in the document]
nomadbl: This suggestion is outdated, you must have commented somehow on an older commit because this line is already in the PR

src/silhouette.jl
```diff
@@ -1,64 +1,100 @@
...
-            r[a[i],j] += dists[i,j]
-        end
-    end
-    return r
+struct SqEuclideanPrecomputeSilhouettes{T} <: PrecomputeSilhouettes
+    nclusters::Int
+    dims::Int
+    counts::Vector{Int} #[nclusters, 1]
+    Ψ::Vector{T} #[nclusters], This represents the second moments of each cluster
+    Y::Matrix{T} #[dims, nclusters], This represents the first moments of each cluster
```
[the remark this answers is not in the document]
nomadbl: The moments are the closest thing there is but its not precisely moments.
For instance these are sums, not averages.

src/silhouette.jl
nomadbl: Here it is actual coordinates.
Is there a reason for the change of places?
If there is, lets see how we can be consistent across all these places

nomadbl:
@alyst 
Thanks for your feedback.
I think we are getting sidetracked with this API thing. I provided it since it looked useful but I now see that we are going to get bogged down in it. It is hard to communicate over github on the API and all the details related to it.
The API I provided is near identical to the one you specified in your detailed comment. Take a look at line 239.
Please lets leave the final API to another PR. This API is completely functional, but you can change it to your liking separately.

I already provided benchmarking exactly as you asked. The two benchmarks compare the `silhouettes()` function with/without precomputation. I don't know how to get the plots, so you can run the benchmark yourself and see them. Here is the text results again for your convenience.
```
"silhouette" => 4-element BenchmarkTools.BenchmarkGroup:
	  tags: []
	  "n=20,000" => 2-element BenchmarkTools.BenchmarkGroup:
		  tags: []
		  "with_precalculation" => Trial(2.102 ms)
		  "without_precalculation" => Trial(4.916 s)
	  "n=10,000" => 2-element BenchmarkTools.BenchmarkGroup:
		  tags: []
		  "with_precalculation" => Trial(1.027 ms)
		  "without_precalculation" => Trial(913.840 ms)
	  "n=1,000" => 2-element BenchmarkTools.BenchmarkGroup:
		  tags: []
		  "with_precalculation" => Trial(91.534 μs)
		  "without_precalculation" => Trial(3.986 ms)
	  "n=100" => 2-element BenchmarkTools.BenchmarkGroup:
		  tags: []
		  "with_precalculation" => Trial(10.111 μs)
		  "without_precalculation" => Trial(42.006 μs)
```

nomadbl:
Another important thing I forgot to write just now:

The cached workflow is the main way I am using this in my code.
The reason, as I alluded to it before, is that frequently the data is so big that it is impractical or impossible to hold it all at once in RAM, so there might not even be a way to call the `silhouettes()` function in practice.
Additionally, I don't see how it is the job of a library to dictate to the user how he should compute things. If there are two ways of doing something, there is no reason to hide one of them.

alyst:
> The API I provided is near identical to the one you specified in your detailed comment. Take a look at line 239.

I have seen it. Yes, it is "near identical" in the sense that it has all the parameters, but in different order, and different parameters are positional vs keyword.
From the perspective of the user of *Clustering.jl* that would mean that the user would have no intuition how functions are organized, and should always refer to the manual.
The particular order is also not just a matter of taste: the input data are provided as positional arguments, method parameters are keywords with the default values corresponding to the most common use case.

Again, I very much welcome your contribution of this improvement to the community.
I just assume that you are also committed to put a bit of effort to integrate it into the existing infrastructure for the benefit of the community.
I am sorry that it means extra work for you, but I have mentioned this issue in the first review.

> I already provided benchmarking exactly as you asked. The two benchmarks compare the `silhouettes()` function with/without precomputation. I don't know how to get the plots, so you can run the benchmark yourself and see them. Here is the text results again for your convenience.

Thank you, the improvement looks impressive.

> The cached workflow is the main way I am using this in my code.
> The reason, as I alluded to it before, is that frequently the data is so big that it is impractical or impossible to hold it all at once in RAM, so there might not even be a way to call the silhouettes() function in practice.
> Additionally, I don't see how it is the job of a library to dictate to the user how he should compute things. If there are two ways of doing something, there is no reason to hide one of them.

Exporting and advertising certain API in the documentation means that the maintainers are committed to maintain both the functionality and the API.
Breaking API compatibility between the versions is painful both for the users and the maintainers.
I hope we both agree that enabling distributed processing is a big thing, which requires thorough consideration.
I personally would not do it before consulting with the other maintainers.
I also totally understand that you don't want to resolve all these API issues -- that why I proposed to not export this functionality in this PR.
But you can still use non-exported functions and types from your code.
For example, we were using `Base.permute!!()`, which was never exported in Julia Base, and recently we had to remove it, because Julia devs decided that it brings no benefits (#229).
Had `permute!!()` been exported, Julia devs would have to wait until 2.0 to break the API.

nomadbl:
Thank you for your detailed explanations and comments. It is very much appreciated.
I adjusted the API according to your comment.
Additionally, the cache struct and function are not exported. I am totally fine with this for my usage, but I absolutely think it should be exported and properly documented once the API is ironed out. I'll help as much as I can with this. I think a discussion in another medium of communication would be helpful.

nomadbl:
@alyst 
I think I addressed your concerns with the last commits, but let me know if I missed anything please.
Also, I understand if you don't have more time for this, so if that is the case kindly offer an additional reviewer.
Your efforts are very much appreciated in either case!

alyst:
@nomadbl Sorry, I didn't have time to finalize this PR. It's on my agenda, I hope to do it over the weekend. There were some comments regarding how the code is divided into functions that were not completely addressed, but I figured out that instead of another review round, it would be easier for both of us if I just commit them myself. Anyway, thanks for your contribution, it is a very nice improvement for silhouettes!

alyst (approved):
After having a closer look at the PR, I've realized that `SilhouettesCache` actually does not do anything silhouette-specific. It is essentially a method for optimized cluster-to-point distances. So I've refactored the code renaming the `SilhouettesCache` into `ClusterDistance` and moving it to the new `cluster_distance.jl` source file. That way it is more visible and could be reused for other algorithms. The "classic" distance matrix-based approach is handled by the `SimpleClusterDistance` class. It also handles any metrics that don't have a dedicated implementation.

I have committed my changes directly to your branch.
There were also some changes to master branch to fix the CI for the new version of *Distances.jl*, so I had to rebase this PR and force-push it to fix the CI.
You would need a fresh local check-out if you want to continue working on it.

So please take a look at my changes. Unfortunately, it was a bit more than I have anticipated, but it was required to make sure that *Clustering.jl* codebase remains consistent.
Let me know if you have any comments. If you are fine, I'll merge the PR. 

Thanks again for your work, the benchmarks show some dramatic improvements for SqEuclidean.
I think after this refactoring implementing #258 (`CosineClusterDistance`) should be easier.

nomadbl:
@alyst 
Thanks a lot!
I'll give this a proper look soon and see if there's any specific comments.
A preliminary comment/clarification is that the method doesn't store distances, and it does not compute them either. What it does do is cache quantities used to compute *sums* of distances. So for example, here, the sums of distances of points in sum cluster to another point.
But yes, I do agree on the general point that this can be possibly used anywhere such distance sums are used, to compute them more efficiently (time and memory wise), so it is a good idea to make a separate file for it.

Again, I'll take a closer look soon, and thanks for your help and input!

alyst:
@nomadbl Did you have a change to look at the PR?

nomadbl:
Sorry about that, I'll take a good look tomorrow.

src/cluster_distances.jl
```diff
@@ -0,0 +1,233 @@
...
+# for SqEuclidean point distances
+struct SqEuclideanClusterDistances{T} <: ClusterDistances{T}
+    npoints::Int
+    cluster_sizes::Vector{Int} # [nclusters]
+    Y::Matrix{T} # [dims, nclusters], the first moments of each cluster (sum of point coords)
+    Ψ::Vector{T} # [nclusters], the second moments of each cluster (sum of point coord squares)
+
+    SqEuclideanClusterDistances(::Type{T}, npoints::Integer, dims::Integer,
+                                nclusters::Integer) where T<:Real =
+        new{T}(npoints, zeros(Int, nclusters), zeros(T, dims, nclusters), zeros(T, nclusters))
```
nomadbl, line 198: `npoints` is set here and is never updated during `update!`
This is fine for small datasets, but imagine cases of
1. a streaming data source where the number of points is not known in advance. In particular the points used for calculating silhouettes is not the same as the number of points that were used for training.
2. The dataset is so big that we can actually calculate in a streaming fashion until silhouette scores converge, and we stop the calculation for actual `npoints` less than this predefined value.

In short, `npoints` should be updated along with `cluster_sizes` on line 218.

src/cluster_distances.jl
```diff
@@ -0,0 +1,233 @@
...
+end
+
+ClusterDistances(::Type{T}, metric::SqEuclidean, npoints::Integer, dims::Integer, nclusters::Integer) where T<:Real =
+    SqEuclideanClusterDistances(T, npoints, dims, nclusters)
+
+Base.isready(dists::SqEuclideanClusterDistances) = sum(dists.cluster_sizes) == dists.npoints
```
nomadbl, line 204: If we apply my suggestion above this method is not necessary

nomadbl, line 204: If we apply my suggestion above this method is not necessary

nomadbl (commented):
Looks great and honestly hard to tell the difference from the previous version aside from some naming changes. But I suppose the bigger changes are actually on the side of the "usual" calculation?
The suggestion I made is small but potentially important, so I'd appreciate it.

Again, sorry for the delay with this review.

alyst:
@nomadbl 
> Looks great and honestly hard to tell the difference from the previous version aside from some naming changes. But I suppose the bigger changes are actually on the side of the "usual" calculation?

There were no changes to the formulas, but some refactorings to avoid code duplication and simplify logic by utilizing Julia method dispatch mechanism.
Calculation of cluster distances (ex-SilhouettesCache) is now separated from `silhouette()` code.

Thank you for the review. I have incorporated your feedback, and if there are no further concerns, I will merge it.

nomadbl:
@alyst
I think merging is fine. It's better if I give improvement suggestions based on actual use from this point.

nomadbl:
@alyst
I think merging is fine. It's better if I give improvement suggestions based on actual use from this point.

alyst:
Released as 0.15.5

PR #257, opened 2023-07-11
Providing intrinsic clustering quality indices

jaksle:
Continuation of issue #253. It introduces 4 indices: Calinsky-Harabasz, Davies-Bouldin, Xie-Beni and Dunn. CH, DB, XB require that the space is linear, e.g. we can talk about average positions, k-means and fuzzy c-means work under the same assumptions. Only CH and XB are defined for fuzzy clustering. D is the most computationally complex, but is the most general, only distances between data points are required, as such works also for k-medoids.

Issues:
- No documentation and testing. Please first check if my API fits Clustering.jl and I will add it. It may be also reasonable to introduce function qualityindex(method,args), maybe not even export the rest of the functions? I am not sure.
- File mytest.jl is temporary testing unit for the current API, will be removed later.
- These methods do not work for arrays which are not 1-idexed. However, the rest of Clustering.jl also does not. Should we add Base.require_one_based_indexing? Theoretically it can be made indexing independent, but it seems irritating to implement.


alyst (changes requested):
Thanks for your contribution!
These metrics (I find the term *index* a bit confusing in this context) are indeed very useful, but the code needs some work.
I've indicated in the code suggestions that we need to export the single function *clustering_quality()*.
I've just reviewed the first `calinski_harabasz()` method, but for the other ones please update the formatting, variable names and linear algebra code in the similar way.
Also please add the proper unit tests (you have them already, but they are in the wrong source file)

alyst:
> These methods do not work for arrays which are not 1-idexed. However, the rest of Clustering.jl also does not. Should we add Base.require_one_based_indexing? Theoretically it can be made indexing independent, but it seems irritating to implement.

Let's try to make the new code in *Clustering.jl* index base-agnostic. It should not be so "irritating": using *eachindex()/eachslice()* should address most of the issues (and potentially even simplify & speed-up the code).

src/Clustering.jl
```diff
@@ -68,7 +68,10 @@ module Clustering
     mcl, MCLResult,
 
     # pair confusion matrix
-    confusion
+    confusion,
+
+    # quality indices
+    calinski_harabasz, davies_bouldin, xie_beni, dunn
```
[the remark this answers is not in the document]
jaksle: One question: do we prefer using method symbols or use dispatch over chosen method in a way similar to what function `sort` does?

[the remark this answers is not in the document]
alyst: I agree that dispatching on `ClusteringQualityIndex` type would be more elegant from engineering perspective.
But I think using symbols is a somewhat simpler solution that we already use to e.g. specify the linkage function or branch order in hierarchical clustering.
I would suggest to start with it.

alyst:
> The name. I am not a big fan of "quality index" either, but it seems to be standard. It is used on Wikipedia and in the review articles I found. I don't have a strong opinion here.

If there's a consensus to use *index* in the literature, let's call it *index* (*quality_index* kwarg in *clustering_quality()* function).
One advantage is that it disambiguates clustering quality *index* from distance *metric*. 

> I was thinking about it a little before and the "irritating" (or maybe better "confusing") part was the assignments vector. Generally its elements can be any type I think?

The elements should be of `Integer` type. The *Clustreing.jl* methods allow taking `assignments::AbstractVector{<:Integer}` that could be constructed by the user.
But you are right that most of the methods assume the elements have to be cluster indices from 1 to K without gaps (the *ClusteringResult.assignments* generated by the *Clustering.jl* clustering methods).

src/qualityindices.jl
```diff
@@ -0,0 +1,224 @@
...
+
+function _check_qualityindex_arguments(
+        assignments::AbstractVector{<:Integer}, # assignments (n)
+        dist::AbstractMatrix{<:Real}            # data distance matrix (n x n)
+    )
+    n, m = size(dist)
+    na = length(assignments)
+    n == m || throw(ArgumentError("Distance matrix must be square."))
+    n == na || throw(DimensionMismatch("Inconsistent array dimensions for distance matrix and assignments."))
+end
```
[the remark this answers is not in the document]
jaksle: I cannot find it in utils.jl and the code was modified to be indexing-agnostic anyway. This code had been moved to clustering_quality implementation.

JoFeG:
Hello everyone, 

I'm currently working on intrinsic cluster validity _indices_ and wanted to propose a new feature for the package, but I found this thread open that is more or less what I was thinking.

The development seems to have paused a couple of months ago; I would like to contribute.

Before diving in, I would like to know if help is indeed needed here?

jaksle:
Hi! The current state is that the commit is fully functional and I am waiting for more detailed re-review by @alyst . You can check the files, few most popular quality measures are implemented there.

alyst:
@jaksle oh, sorry, I have missed your updates. I'll try to review them in the next few days. I have recently merged #255, which, I think, put your PR into a conflicting state. Could you please resolve the conflicts?

jaksle:
Conflict resolved. The code is mostly index-agnostic; I learned that `Distances.jl` will not work in the middle of doing it. Other than that, the documentation may be to verbose, but it should be easy to optionally trim down.

src/clustering_quality.jl
```diff
@@ -0,0 +1,328 @@
...
+
+    min_outer_distance = minimum(
+        minimum(view(dist, cluster_samples[j₁], cluster_samples[j₂]), init = typemax(eltype(dist)))
+            for j₁ in 1:k for j₂ in j₁+1:k
+    )
+
+    max_inner_distance = maximum(
+        maximum(dist[i₁,i₂] for i₁ in sample, i₂ in sample, init = typemin(eltype(dist)))
+            for sample in cluster_samples
+    )
```
alyst, line 325: Using `assignments` directly looks faster and does not allocate:
```suggestion
    max_inner_distance = typemin(eltype(dist))
    min_outer_distance = typemax(eltype(dist))
    for ((i, clu_i), dist_i) in zip(enumerate(assignments), eachcol(dist))
        for j in (i + 1):k
            dist = dist_i[j]
            if assignments[j] == clu_i
                if max_inner_distance > dist
                    max_inner_distance = dist
                end
            else
                if min_outer_distance < dist
                    min_outer_distance = dist
                end
            end
        end
    end
```

test/clustering_quality.jl
```diff
@@ -0,0 +1,58 @@
...
+    # using Plots
+    # scatter(Y[:,1],Y[:,2],
+    #     axisratio = :equal,
+    #     #seriescolor = palette(default)[A],
+    # )
+    # scatter!(C[:,1],C[:,2],
+    #     marker = :square,
+    #     label = "cluster centers",  
+    # )
+
```
alyst, line 37: please remove

test/clustering_quality.jl
```diff
@@ -0,0 +1,58 @@
...
+    @testset "input checks" begin
```
alyst, line 38: can you test for incorrect `quality_index` for each of the `clustering_quality()` variant here?

alyst (changes requested):
Thank you, there was a big progress since the last version!
I've left some comments/suggestions to the docs and the code.
I have not made these suggestions everywhere, but please use space after comma in the code and in the docs.

At the moment `clustering_quality.jl` tests are not enabled in the CI, so it is not clear whether the code runs and gives the correct results, and what is the coverage. Please include this file in the `runtests.jl`. By looking at the existing tests I would say that ideally we would also need the tests with non-0,1 weights and with varying fuzzyness.

If the CI results look ok, I think we are close to merging it after the final review.

docs/source/validate.md
```diff
@@ -54,6 +54,147 @@ Higher values indicate better separation of clusters w.r.t. point distances.
...
+```
+### [Dunn index](https://en.wikipedia.org/wiki/Dunn_index) (↑) 
+Option `:dunn`. More computationally demanding index which can be used when the centres are not known. It measures ratio between the nearest neighbour distance divided by the maximum cluster diameter. It is defined as
+```math
+\frac{\min\limits_{ C_{j_1}\neq C_{j_2}} \delta(C_{j_1},C_{j_2})}{\max\limits_{C_j}\Delta(C_j)}
+```
+
+where
+```math
+\delta(C_{j_1},C_{j_2}) = \min\limits_{x_{i_1}\in C_{j_1},x_{i_2}\in C_{j_2}} d(x_{i_1},x_{i_2}),\quad \Delta(C_j) = \max\limits_{x_{i_1},x_{i_2}\in C_j} d(x_{i_1},x_{i_2}).
```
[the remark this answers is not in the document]
alyst: Ah, and `\delta(C_{j_1},C_{j_2})` into `\mathrm{distance}(C_{j_1}, C_{j_2})`

src/clustering_quality.jl
```diff
@@ -0,0 +1,328 @@
...
+
+clustering_quality(X::AbstractMatrix{<:Real}, R::FuzzyCMeansResult, fuzziness::Real, distance::SemiMetric=SqEuclidean(); quality_index::Symbol) =
+    clustering_quality(X, R.centers, R.weights, fuzziness, distance; quality_index)
+
+"""
+
+    clustering_quality(assignments, dist_matrix; quality_index)
+    clustering_quality(clustering, dist_matrix; quality_index)
+    clustering_quality(data, assignments, [distance;] quality_index)
+    clustering_quality(data, clustering, [distance;] quality_index)
```
alyst, line 126:
```suggestion
    clustering_quality(data, assignments; quality_index, [metric])
    clustering_quality(data, clustering; quality_index, [metric])
```

test/clustering_quality.jl
```diff
@@ -0,0 +1,58 @@
...
+        @test_throws ArgumentError clustering_quality(zeros(2,2), zeros(2,3), [1, 2], quality_index = :calinski_harabasz)
+        @test_throws DimensionMismatch clustering_quality(zeros(2,2),zeros(3,2), [1, 2], quality_index = :calinski_harabasz)
+        @test_throws ArgumentError clustering_quality(zeros(2,2),zeros(2,1), [1, ], quality_index = :calinski_harabasz)
+        @test_throws ArgumentError clustering_quality(zeros(2,2),zeros(2,2), [1, 2], quality_index = :calinski_harabasz)
+        @test_throws DimensionMismatch clustering_quality([1,2,3], zeros(2,2), quality_index = :dunn)
+    end
+
```
alyst, line 45: We should also test corner cases to make sure they are handled correctly (ideally, the correct value according to the quality index definition should be returned):
  * degenerated input data: 0 points (and 0 clusters)
  * trivial clustering: single cluster
  * 0-dimensional data

jaksle: Quality indices are generally undefined for single cluster. The reason is that the cluster spread is undefined.

alyst: Ok, *silhouettes()* actually throws an error for the degenerated clustering, so the other indices should follow that behaviour.

The other corner case that *silhouettes()* checks are empty clusters (gaps in `assignments` values).
See [test/silhouette.jl](https://github.com/JuliaStats/Clustering.jl/blob/master/test/silhouette.jl).

jaksle: If the cluster is empty than it's center is undefined, which means that the user gave nonsense arguments (or inconsistent number of assignments/centers, this is checked). Only silhouettes and Dunn index do not require them.

src/clustering_quality.jl
alyst: Ah, there's a bug in my code: the j loop should be `for j in (i + 1):n`, where `n = length(assignments)`

alyst: we just iterate over all pairs of points and update max inner distance if the pair is in the same cluster or min outer distance otherwise

test/clustering_quality.jl
alyst: ideally we need to test that the behavior in such cases meets our expectations, because this will happen (bugs in the user code). I'd say: let's add the tests, see how `clustering_quality` handles it  and then decide if we need to adjust the behavior.

src/clustering_quality.jl
```diff
@@ -0,0 +1,328 @@
...
+        assignments::AbstractVector{<:Integer},
+        distance::SemiMetric=SqEuclidean()
+    )
+
+    k = size(centers, 2)
+    c_idx = axes(centers,2)
+
+    cluster_samples = _gather_samples(assignments, k)
+
+    cluster_diameters = [mean(colwise(distance,view(X, :, sample), centers[:,j])) for (j, sample) in zip(c_idx, cluster_samples) ]
```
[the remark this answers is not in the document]
jaksle: I think it is fine now. `cluster_diameters` are $S(C_k)$ from the formula

.gitignore
```diff
@@ -1,2 +1,8 @@
 doc/build
 Manifest.toml
+*.swp
+.vscode/settings.json
+docs/build/
+docs/build/assets/
+Project.toml
+*.html
```
alyst, line 8: removing `Project.toml` is a bad idea.
If we want to ignore VScode files, I'd rather remove the whole folder.
Ignoring `docs/build` makes sense, this rule should cover all subfolders and generated files.

```suggestion
*.swp
.vscode
docs/build/
```

Project.toml
```diff
@@ -4,9 +4,11 @@ version = "0.15.4"
 
 [deps]
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 NearestNeighbors = "b8a86587-4115-5ab1-83bc-aa920d37bbce"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+RDatasets = "ce6b1742-4840-55fa-b093-852dadbb1d8b"
```
alyst, line 11: Please roll these changes back.
Documentation generation uses `docs/Project.toml`, and Documenter is already there.

```suggestion
LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
NearestNeighbors = "b8a86587-4115-5ab1-83bc-aa920d37bbce"
Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
```

docs/source/validate.md
```diff
@@ -54,6 +54,147 @@ Higher values indicate better separation of clusters w.r.t. point distances.
...
+### Calinski-Harabasz index (↑) 
+
+Option `:calinski_harabasz`. Measures corrected ratio between the summed internal inertia of clusters divided by global inertia of the cluster centers. For hard clustering and soft (fuzzy) it is defined as
+
+```math
+
+\frac{n-k}{k-1}\frac{\sum_{C_j}|C_j|d(c_j,c)}{\sum\limits_{C_j}\sum\limits_{x_i\in C_j} d(x_i,c_j)} \quad \text{and}\quad 
+\frac{n-k}{k-1} \frac{\sum_{C_j}\sum_{x_i} w_{ik}^md(x_i,c_j)}{\sum\limits_{C_j}\sum\limits_{x_i}w_{ij}^m d(c_j,c)}
+```
+respectively.
```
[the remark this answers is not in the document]
jaksle: It should be outer / inner. Only the text was wrong, the formulas and code were good. Corrected.

test/clustering_quality.jl
```diff
@@ -0,0 +1,57 @@
...
+    @testset "correct index values" begin
+        @test clustering_quality(Y', C', A; quality_index = :calinski_harabasz, metric = Euclidean()) ≈ (32/3) / (16/8)
+        @test clustering_quality(Y', C', W; quality_index = :calinski_harabasz, fuzziness = 2, metric = Euclidean()) ≈ (32/3) / (16/8)
+
+        @test clustering_quality(Y', C', A; quality_index = :davies_bouldin, metric = Euclidean()) ≈ 3/2√5
```
alyst, line 49: I guess it fails on 1.0 because of `√5`.
Could you try replacing it with sqrt?
```suggestion
        @test clustering_quality(Y', C', A; quality_index = :davies_bouldin, metric = Euclidean()) ≈ 3/2 sqrt(5)
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,317 @@
...
+    end
+
+    if quality_index == :calinski_harabasz
+        _cluquality_calinski_harabasz(data, centers, assignments, metric)
+    elseif quality_index == :xie_beni
+        _cluquality_xie_beni(data, centers, assignments, metric)
+    elseif quality_index == :davies_bouldin
+        _cluquality_davies_bouldin(data, centers, assignments, metric)
+    else quality_index == :davies_bouldin
+    if quality_index == :silhouettes
```
alyst, line 79:
```suggestion
    elseif quality_index == :silhouettes
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,317 @@
...
+        mean(silhouettes(assignments, pairwise(metric, eachcol(data))))
+    elseif quality_index == :dunn 
+        _cluquality_dunn(assignments, pairwise(metric, eachcol(data)))
+    else
+        throw(ArgumentError("Quality index $quality_index not supported."))
+    end
+end
```
alyst, line 86:
```suggestion
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,315 @@
...
+    k == kw || throw(DimensionMismatch("Inconsistent number of clusters for `centers` and `weights`."))
+    (1 <= k <= n) || throw(ArgumentError("Number of clusters k must be from 1:n (n=$n), k=$k given."))
+    k >= 2 || throw(ArgumentError("Quality index not defined for the degenerated clustering with a single cluster."))
+    n == k && throw(ArgumentError("Quality index not defined for the degenerated clustering where each data point is its own cluster."))
+    all(>=(0), weights) || throw(ArgumentError("All weights must be larger or equal 0."))
+    1 < fuzziness || throw(ArgumentError("Fuzziness must be greater than 1 ($fuzziness given)"))
+
+    if quality_index == :calinski_harabasz
+        _cluquality_calinski_harabasz(data, centers, weights, fuzziness, metric)
+    elseif quality_index ∈ (:xie_beni, :Xie_Beni, :xb)
```
alyst, line 116:
```suggestion
    elseif quality_index == :xie_beni
```

test/clustering_quality.jl
```diff
@@ -0,0 +1,57 @@
...
+        @test_throws ArgumentError clustering_quality(zeros(2,2),zeros(2,1), [1, ], quality_index = :calinski_harabasz)
+        @test_throws ArgumentError clustering_quality(zeros(2,2),zeros(2,2), [1, 2], quality_index = :calinski_harabasz)
+        @test_throws DimensionMismatch clustering_quality([1,2,3], zeros(2,2), quality_index = :dunn)
+    end
+
+    @testset "correct index values" begin
+        @test clustering_quality(Y', C', A; quality_index = :calinski_harabasz, metric = Euclidean()) ≈ (32/3) / (16/8)
+        @test clustering_quality(Y', C', W; quality_index = :calinski_harabasz, fuzziness = 2, metric = Euclidean()) ≈ (32/3) / (16/8)
+
+        @test clustering_quality(Y', C', A; quality_index = :davies_bouldin, metric = Euclidean()) ≈ 3/2 sqrt(5)
```
alyst, line 49:
```suggestion
        @test clustering_quality(Y', C', A; quality_index = :davies_bouldin, metric = Euclidean()) ≈ 3/sqrt(20)
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,315 @@
...
+    if quality_index == :calinski_harabasz
+        _cluquality_calinski_harabasz(data, centers, assignments, metric)
+    elseif quality_index == :xie_beni
+        _cluquality_xie_beni(data, centers, assignments, metric)
+    elseif quality_index == :davies_bouldin
+        _cluquality_davies_bouldin(data, centers, assignments, metric)
+    elseif quality_index == :silhouettes
+        mean(silhouettes(assignments, pairwise(metric, eachcol(data))))
+    elseif quality_index == :dunn 
+        _cluquality_dunn(assignments, pairwise(metric, eachcol(data)))
```
alyst, line 81:
```suggestion
        mean(silhouettes(assignments, pairwise(metric, data, dims=2)))
    elseif quality_index == :dunn 
        _cluquality_dunn(assignments, pairwise(metric, data, dims=2))
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,315 @@
...
+        error(ArgumentError("Quality index $quality_index not available."))
+    end
+end
+
+
+clustering_quality(data::AbstractMatrix{<:Real}, assignments::AbstractVector{<:Integer}; quality_index::Symbol, metric::SemiMetric=SqEuclidean()) = 
+    clustering_quality(assignments, pairwise(metric,eachcol(data)); quality_index = quality_index)
+
+clustering_quality(data::AbstractMatrix{<:Real}, R::ClusteringResult;  quality_index::Symbol, metric::SemiMetric=SqEuclidean()) =
+    clustering_quality(R.assignments, pairwise(metric,eachcol(data)); quality_index = quality_index)
```
alyst, line 153:
```suggestion
    clustering_quality(assignments, pairwise(metric, data, dims=2); quality_index = quality_index)

clustering_quality(data::AbstractMatrix{<:Real}, R::ClusteringResult;  quality_index::Symbol, metric::SemiMetric=SqEuclidean()) =
    clustering_quality(R.assignments, pairwise(metric, data, dims=2); quality_index = quality_index)
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,315 @@
...
+    inner_inertia = sum(
+        sum(colwise(metric, view(data, :, samples), center))
+            for (center, samples) in zip(eachcol(centers), cluster_samples)
+    )
+    return inner_inertia
+end
+
+function _inner_inertia(data, centers, weights, fuzziness, metric) # shared between fuzzy clustering calinski_harabasz and xie_beni
+
+    pointCentreDistances = pairwise(metric, eachcol(data), eachcol(centers))
```
alyst, line 180:
```suggestion
    pointCentreDistances = pairwise(metric, data, centers, dims=2)
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,315 @@
...
+        weights::AbstractMatrix{<:Real},
+        fuzziness::Real,
+        metric::SemiMetric=SqEuclidean()
+    )
+
+    n, k = size(data, 2), size(centers, 2)
+
+    inner_intertia = _inner_inertia(data, centers, weights, fuzziness, metric)
+
+    center_distances = pairwise(metric, eachcol(centers))
```
alyst, line 289:
```suggestion
    center_distances = pairwise(metric, centers, dims=2)
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,315 @@
...
+function _inner_inertia(data, centers, weights, fuzziness, metric) # shared between fuzzy clustering calinski_harabasz and xie_beni
+
+    pointCentreDistances = pairwise(metric, data, centers, dims=2)
+
+    inner_inertia = sum(
+        w^fuzziness * d for (w, d) in zip(weights, pointCentreDistances) 
+    )
+
+    return inner_inertia
+end
```
alyst, line 187:
```suggestion
# shared between fuzzy clustering calinski_harabasz and xie_beni (fuzzy version)
function _inner_inertia(metric::SemiMetric, data::AbstractMatrix, centers::AbstractMatrix,
                        weights::AbstractMatrix, fuzziness::Real)
    pointCentreDistances = pairwise(metric, data, centers, dims=2)
    inner_inertia = sum(
        w^fuzziness * d for (w, d) in zip(weights, pointCentreDistances) 
    )
    return inner_inertia
end
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,315 @@
...
+    )
+
+    n, k = size(data, 2), size(centers, 2)
+
+    cluster_samples = _gather_samples(assignments, k)
+    global_center = vec(mean(data, dims=2))
+    center_distances = colwise(metric, centers, global_center)
+    outer_inertia = length.(cluster_samples) ⋅ center_distances
+
+    inner_inertia = _inner_inertia(data, centers, cluster_samples, metric)
```
alyst, line 204:
```suggestion
    global_center = vec(mean(data, dims=2))
    center_distances = colwise(metric, centers, global_center)
    outer_inertia = sum(center_distances[clu] for clu in assignments)
    inner_inertia = _inner_inertia(metric, data, centers, assignments)
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,315 @@
...
+end
+
+
+function _inner_inertia(data, centers, cluster_samples, metric) # shared between hard clustering calinski_harabasz and xie_beni
+    inner_inertia = sum(
+        sum(colwise(metric, view(data, :, samples), center))
+            for (center, samples) in zip(eachcol(centers), cluster_samples)
+    )
+    return inner_inertia
+end
```
alyst, line 176:
```suggestion
# shared between hard clustering calinski_harabasz and xie_beni
_inner_inertia(metric::SemiMetric, data::AbstractMatrix, centers::AbstractMatrix,
               assignments::AbstractVector{<:Integer}) =
    sum(metric(view(data, :, i), view(center, :, clu)) for (i, clu) in enumerate(assignments))
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,315 @@
...
+    n, k = size(data, 2), size(centers, 2)
+
+    global_center = vec(mean(data, dims=2))
+    center_distances = colwise(metric, centers, global_center)
+
+    outer_inertia = 
+        sum(sum(w^fuzziness for w in w_col) * d
+            for (w_col, d) in zip(eachcol(weights), center_distances)
+        )
+    inner_inertia = _inner_inertia(data, centers, weights, fuzziness, metric)
```
alyst, line 226:
```suggestion
    outer_inertia = sum(sum(w^fuzziness for w in view(weights, :, clu)) * d
            for (clu, d) in enumerate(center_distances)
        )
    inner_inertia = _inner_inertia(metric, data, centers, weights, fuzziness)
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,315 @@
...
+        data::AbstractMatrix{<:Real},
+        centers::AbstractMatrix{<:Real},
+        assignments::AbstractVector{<:Integer},
+        metric::SemiMetric=SqEuclidean()
+    )
+
+    n, k = size(data, 2), size(centers,2)
+
+    cluster_samples = _gather_samples(assignments, k)
+    inner_intertia  = _inner_inertia(data, centers, cluster_samples, metric)
```
alyst, line 268:
```suggestion
    inner_intertia  = _inner_inertia(metric, data, centers, assignments)
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,315 @@
...
+
+    center_distances = pairwise(metric,centers)
+    min_center_distance = minimum(center_distances[j₁,j₂] for j₁ in 1:k for j₂ in j₁+1:k)
```
alyst, line 271:
```suggestion
    inner_intertia  = _inner_inertia(metric, data, centers, assignments)
    center_distances = pairwise(metric, centers, dims=2)
    min_center_distance = minimum(center_distances[j₁,j₂] for j₁ in 1:k for j₂ in j₁+1:k)
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,315 @@
...
+        fuzziness::Real,
+        metric::SemiMetric=SqEuclidean()
+    )
+
+    n, k = size(data, 2), size(centers, 2)
+
+    inner_intertia = _inner_inertia(data, centers, weights, fuzziness, metric)
+
+    center_distances = pairwise(metric, centers, dims=2)
+    min_center_distance = minimum(center_distances[j₁,j₂] for j₁ in 1:k for j₂ in j₁+1:k)
```
alyst, line 289:
```suggestion
    inner_intertia = _inner_inertia(metric, data, centers, weights, fuzziness)
    center_distances = pairwise(metric, centers, dims=2)
    min_center_distance = minimum(center_distances[j₁,j₂] for j₁ in 1:k for j₂ in j₁+1:k)
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,304 @@
...
+        push!(cluster_samples[a], i)
+    end
+    return cluster_samples
+end
+
+
+# shared between calinski_harabasz and xie_beni (hard version)
+_inner_inertia(metric::SemiMetric, data::AbstractMatrix, centers::AbstractMatrix,
+               assignments::AbstractVector{<:Integer}) =
+    sum(metric(view(data, :, i), view(center, :, clu)) for (i, clu) in enumerate(assignments))
```
alyst, line 173:
```suggestion
    sum(metric(view(data, :, i), view(centers, :, clu)) for (i, clu) in enumerate(assignments))
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,304 @@
...
+    for (i, a) in zip(eachindex(assignments), assignments)
+        push!(cluster_samples[a], i)
+    end
+    return cluster_samples
+end
+
+
+# shared between calinski_harabasz and xie_beni (hard version)
+_inner_inertia(metric::SemiMetric, data::AbstractMatrix, centers::AbstractMatrix,
+               assignments::AbstractVector{<:Integer}) =
```
alyst, line 172:
```suggestion
_inner_inertia(metric::SemiMetric, data::AbstractMatrix, centers::AbstractMatrix,
               assignments::AbstractVector{<:Integer}, fuzziness::Nothing) =
```

jaksle: You can calculate inner_inertia without using `_gather_samples` as you suggest, but `colwise` can be more efficient than calling chosen distance within the loop. This makes possible gains unclear and distance-dependent.

src/clustering_quality.jl
```diff
@@ -0,0 +1,304 @@
...
+
+    global_center = vec(mean(data, dims=2))
+    center_distances = colwise(metric, centers, global_center)
+
+    outer_inertia = sum(sum(w^fuzziness for w in view(weights, :, clu)) * d
+                        for (clu, d) in enumerate(center_distances))
+    inner_inertia = _inner_inertia(metric, data, centers, weights, fuzziness)
+
+    return (outer_inertia / inner_inertia) * (n - k) / (k - 1)
+end
```
alyst, line 222:
```suggestion
# "hard" version
function _outer_inertia(metric::SemiMetric, data::AbstractMatrix, centers::AbstractMatrix,
                        assignments::AbstractVector{<:Integer}, fuzziness::Nothing)
    global_center = vec(mean(data, dims=2))
    center_distances = colwise(metric, centers, global_center)
    return sum(center_distances[clu] for clu in assignments)
end

# "fuzzy" version
function _outer_inertia(metric::SemiMetric, data::AbstractMatrix, centers::AbstractMatrix,
                        weights::AbstractMatrix, fuzziness::Real)
    global_center = vec(mean(data, dims=2))
    center_distances = colwise(metric, centers, global_center)
    return sum(sum(w^fuzziness for w in view(weights, :, clu)) * d
               for (clu, d) in enumerate(center_distances))
end

function  _cluquality_calinski_harabasz(
        metric::SemiMetric,
        data::AbstractMatrix{<:Real},
        centers::AbstractMatrix{<:Real},
        assignments::Union{AbstractVector{<:Integer}, AbstractMatrix{<:Real}},
        fuzziness::Union{Real, Nothing}
)
    n, k = size(data, 2), size(centers, 2)
    outer_inertia = _outer_inertia(metric, data, centers, assignments, fuzziness)
    inner_inertia = _inner_inertia(metric, data, centers, assignments, fuzziness)
    return (outer_inertia / inner_inertia) * (n - k) / (k - 1)
end
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,304 @@
...
+    )
+
+    n, k = size(data, 2), size(centers, 2)
+
+    inner_intertia = _inner_inertia(metric, data, centers, weights, fuzziness)
+    center_distances = pairwise(metric, centers, dims=2)
+    min_center_distance = minimum(center_distances[j₁,j₂] for j₁ in 1:k for j₂ in j₁+1:k)
+
+    return inner_intertia / (n * min_center_distance)
+end
```
alyst, line 282:
```suggestion
function _cluquality_xie_beni(
        metric::SemiMetric,
        data::AbstractMatrix{<:Real},
        centers::AbstractMatrix{<:Real},
        assignments::Union{AbstractVector{<:Integer}, AbstractMatrix{<:Real}},
        fuzziness::Union{Real, Nothing}
)
    n, k = size(data, 2), size(centers,2)
    inner_intertia  = _inner_inertia(metric, data, centers, assignments, fuzziness)
    center_distances = pairwise(metric, centers, dims=2)
    min_center_distance = minimum(center_distances[j₁,j₂] for j₁ in 1:k for j₂ in j₁+1:k)
    return inner_intertia / (n * min_center_distance)
end
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,304 @@
...
+    (1 <= k <= n) || throw(ArgumentError("Number of clusters k must be from 1:n (n=$n), k=$k given."))
+    k >= 2 || throw(ArgumentError("Quality index not defined for the degenerated clustering with a single cluster."))
+    n == k && throw(ArgumentError("Quality index not defined for the degenerated clustering where each data point is its own cluster."))
+    all(>=(0), weights) || throw(ArgumentError("All weights must be larger or equal 0."))
+    1 < fuzziness || throw(ArgumentError("Fuzziness must be greater than 1 ($fuzziness given)"))
+
+    if quality_index == :calinski_harabasz
+        _cluquality_calinski_harabasz(data, centers, weights, fuzziness, metric)
+    elseif quality_index == :xie_beni
+        _cluquality_xie_beni(data, centers, weights, fuzziness, metric)
```
alyst, line 117:
```suggestion
        _cluquality_calinski_harabasz(metric, data, centers, weights, fuzziness)
    elseif quality_index == :xie_beni
        _cluquality_xie_beni(metric, data, centers, weights, fuzziness)
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,304 @@
...
+    k >= 2 || throw(ArgumentError("Quality index not defined for the degenerated clustering with a single cluster."))
+    n == k && throw(ArgumentError("Quality index not defined for the degenerated clustering where each data point is its own cluster."))
+    for i in eachindex(assignments)
+        (assignments[i] in axes(centers, 2)) || throw(ArgumentError("Bad assignments[$i]=$(assignments[i]) is not a valid index for `data`."))
+    end
+
+    if quality_index == :calinski_harabasz
+        _cluquality_calinski_harabasz(data, centers, assignments, metric)
+    elseif quality_index == :xie_beni
+        _cluquality_xie_beni(data, centers, assignments, metric)
```
alyst, line 75:
```suggestion
        _cluquality_calinski_harabasz(metric, data, centers, assignments, nothing)
    elseif quality_index == :xie_beni
        _cluquality_xie_beni(metric, data, centers, assignments, nothing)
```

src/clustering_quality.jl
alyst: `pairwise()` is for sure more efficient when we are calculating pairwise distances (for certain metrics) between N and M points.
But in this case it's a distance between N points and 1 point (in `_gather_samples()` implementation too).
Is it also getting significant performance boost?

alyst:
@jaksle We have reached the point when the CI tests are passing, thank you!
If we look at the coverage report (the first PR comment), not all the lines are covered, though.
In particular, testing for incorrect :quality_index values, and some variants of the `clustering_quality()` signatures are never called.
Could you please add tests to cover those lines?

src/clustering_quality.jl
jaksle: For what I understand, possible gains are entirely metric-dependent. For the most popular one, in
```julia
using BenchmarkTools

X = randn(1000,1000)
C = randn(1000)

@benchmark sum(colwise(Euclidean(),X,C))
@benchmark sum(Euclidean()(x,C) for x in eachcol(X))
```
on my machine I get `colwise` being faster by 60% and taking 10x less memory.

alyst: Thanks for checking. Could you please add your benchmark results here?
I guess the performance gain is due to the use of BLAS matrix multiplication for Euclidean/SqEuclidean (so for the other metrics the gain should be lower).
The memory inefficiency is a bit unfortunate, I'm not sure it's because of `eachcol(X)` vs `view(X, :, i)`, but due to internal arrays allocation in the generic `metric()` code.
Anyway, thanks again for checking -- you can revert it to the old version (otherwise I will do it), just keep the signature of `inner_inertia()` the same and call `gather_samples()` inside it.

src/clustering_quality.jl
```diff
@@ -0,0 +1,274 @@
...
+
+
+# Dunn index
+
+function _cluquality_dunn(assignments::AbstractVector{<:Integer}, dist::AbstractMatrix{<:Real})
+
+    max_inner_distance, min_outer_distance = typemin(eltype(dist)), typemax(eltype(dist))
+    
+    for i in eachindex(assignments), j in (i + 1):lastindex(assignments)
+        d = dist[i,j]
```
alyst, line 262:
```suggestion
function _cluquality_dunn(dist::AbstractMatrix{<:Real}, assignments::AbstractVector{<:Integer})
    T = eltype(dist)
    max_inner_distance, min_outer_distance = typemin(T), typemax(T)
    
    for i in eachindex(assignments), j in (i + 1):lastindex(assignments)
        @inbounds d = dist[i,j]
```

alyst, line 262:
```suggestion
function _cluquality_dunn(dist::AbstractMatrix{<:Real}, assignments::AbstractVector{<:Integer})
    T = eltype(dist)
    max_inner_distance, min_outer_distance = typemin(T), typemax(T)

    for i in eachindex(assignments), j in (i + 1):lastindex(assignments)
        @inbounds d = dist[i,j]
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,274 @@
...
+    )
+    n, m = size(dist)
+    na = length(assignments)
+    n == m || throw(ArgumentError("Distance matrix must be square."))
+    n == na || throw(DimensionMismatch("Inconsistent array dimensions for distance matrix and assignments."))
+
+    if quality_index == :silhouettes
+        mean(silhouettes(assignments, dist))
+    elseif quality_index == :dunn
+        _cluquality_dunn(assignments, dist)
```
alyst, line 142:
```suggestion
        _cluquality_dunn(dist, assignments)
```

src/clustering_quality.jl
jaksle: Sure, `colwise` is

```julia
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  132.700 μs …  4.223 ms  ┊ GC (min … max): 0.00% … 96.37%
 Time  (median):     137.900 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   144.589 μs ± 47.773 μs  ┊ GC (mean ± σ):  0.28% ±  0.96%

  ▃▇█▆▅▅▃▃▂▂▁▁▁    ▁    ▁                                      ▂
  ██████████████████████████▇▇▆▆▆▇▆▆▆▆▆▅▃▅▅▄▁▄▄▁▄▅▃▁▄▄▃▁▃▄▄▁▅▅ █
  133 μs        Histogram: log(frequency) by time       256 μs <

 Memory estimate: 7.95 KiB, allocs estimate: 2.
```

`Euclidean()(x,C)` is
```julia
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  197.400 μs …   3.395 ms  ┊ GC (min … max): 0.00% … 87.95%
 Time  (median):     203.200 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   211.891 μs ± 122.278 μs  ┊ GC (mean ± σ):  2.17% ±  3.57%

     ▃█▇▇▁         
  ▁▂▅█████▆▄▃▂▂▂▂▂▂▂▂▂▂▂▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  197 μs           Histogram: frequency by time          254 μs <

 Memory estimate: 78.17 KiB, allocs estimate: 3001.
```

I've checked `view(X,:,i)` quickly, it is even worse.

src/clustering_quality.jl
```diff
@@ -0,0 +1,272 @@
...
+
+    cluster_diameters = [mean(colwise(metric,view(data, :, sample), centers[:,j])) for (j, sample) in zip(c_idx, cluster_samples) ]
+    center_distances = pairwise(metric,centers)
+
+    DB = mean(
+        maximum( (cluster_diameters[j₁] + cluster_diameters[j₂]) / center_distances[j₁,j₂] for j₂ in c_idx if j₂ ≠ j₁)
+            for j₁ in c_idx
+    )
+    return  DB
+end
```
alyst, line 237:
```suggestion
function _cluquality_davies_bouldin(
        metric::SemiMetric,
        data::AbstractMatrix{<:Real},
        centers::AbstractMatrix{<:Real},
        assignments::AbstractVector{<:Integer},
)
    clu_idx = axes(centers, 2)
    clu_samples = _gather_samples(assignments, length(clu_idx))
    clu_diams = [mean(colwise(metric, view(data, :, samples), view(centers, :, clu)))
                 for (clu, samples) in zip(clu_idx, clu_samples)]
    center_dists = pairwise(metric, centers, dims=2)

    return mean(maximum((clu_diams[j₁] + clu_diams[j₂]) / center_dists[j₁, j₂]
                        for j₂ in clu_idx if j₂ ≠ j₁)
                for j₁ in clu_idx)
end
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,272 @@
...
+    for i in eachindex(assignments)
+        (assignments[i] in axes(centers, 2)) || throw(ArgumentError("Bad assignments[$i]=$(assignments[i]) is not a valid index for `data`."))
+    end
+
+    if quality_index == :calinski_harabasz
+        _cluquality_calinski_harabasz(metric, data, centers, assignments, nothing)
+    elseif quality_index == :xie_beni
+        _cluquality_xie_beni(metric, data, centers, assignments, nothing)
+    elseif quality_index == :davies_bouldin
+        _cluquality_davies_bouldin(data, centers, assignments, metric)
```
alyst, line 77:
```suggestion
        _cluquality_davies_bouldin(metric, data, centers, assignments)
```

src/clustering_quality.jl
alyst: > I've checked view(X,:,i) quickly, it is even worse.

Worse in terms of memory or performance?

src/clustering_quality.jl
```diff
@@ -0,0 +1,265 @@
...
+)
+    clu_idx = axes(centers, 2)
+    clu_samples = _gather_samples(assignments, length(clu_idx))
+    clu_diams = [mean(colwise(metric, view(data, :, samples), view(centers, :, clu)))
+                 for (clu, samples) in zip(clu_idx, clu_samples)]
+    center_dists = pairwise(metric, centers, dims=2)
+
+    return mean(maximum((clu_diams[j₁] + clu_diams[j₂]) / center_dists[j₁, j₂]
+                        for j₂ in clu_idx if j₂ ≠ j₁)
+                for j₁ in clu_idx)
```
alyst, line 229:
```suggestion
    return mean(maximum(@inbounds (clu_diams[j₁] + clu_diams[j₂]) / center_dists[j₁, j₂]
                        for j₂ in clu_idx if j₂ ≠ j₁)
                for j₁ in clu_idx)
```

src/clustering_quality.jl
jaksle: On the same machine
```julia
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  325.000 μs …  2.099 ms  ┊ GC (min … max): 0.00% … 69.74%
 Time  (median):     339.600 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   348.629 μs ± 85.286 μs  ┊ GC (mean ± σ):  1.09% ±  3.81%

     ▁▆█▆▅▄▃▁▁▂▂▁   ▁                                          ▂
  ▇▃▁█████████████████▇▇▇▇▆▆▅▅▅▅▅▆▅▅▅▄▅▅▁▅▄▃▃▃▄▅▄▁▁▃▃▁▄▃▃▄▄▁▁▄ █
  325 μs        Histogram: log(frequency) by time       478 μs <

 Memory estimate: 116.66 KiB, allocs estimate: 5466.
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,316 @@
...
+    )
+
+    n, k = size(data, 2), size(centers, 2)
+
+    inner_intertia = _inner_inertia(metric, data, centers, weights, fuzziness)
+
+    center_distances = pairwise(metric, centers, dims=2)
+    min_center_distance = minimum(center_distances[j₁,j₂] for j₁ in 1:k for j₂ in j₁+1:k)
+    return inner_intertia / (n * min_center_distance)
+end
```
alyst, line 296: We don't need this method: the code is exactly the same as the one above:
 * for hard clustering `assignments` is the vector of integer cluster indices for each sample, and `fuzziness = nothing`
 * for fuzzy clustering `assignments` is the matrix of sample-to-cluster weights, and `fuzziness` is a real number
 
 So depending on the type of `assignments` and `fuzziness` the correct version of the `inner_inertia` is called.

src/clustering_quality.jl
```diff
@@ -0,0 +1,316 @@
...
+    return cluster_samples
+end
+
+# shared between hard clustering calinski_harabasz and xie_beni
+function _inner_inertia(
+        metric::SemiMetric,
+        data::AbstractMatrix,
+        centers::AbstractMatrix,
+        assignments::AbstractVector{<:Integer}
+    )
```
alyst, line 178: To support clustering type-invariant version of `_cluquality_xie_beny()`
```suggestion
        assignments::AbstractVector{<:Integer},
        fuzziness::Nothing
    ) 
```

src/clustering_quality.jl
```diff
@@ -0,0 +1,316 @@
...
+    n, k = size(data, 2), size(centers, 2)
+
+    global_center = vec(mean(data, dims=2))
+    center_distances = colwise(metric, centers, global_center)
+
+    outer_inertia = 
+        sum(sum(w^fuzziness for w in w_col) * d
+            for (w_col, d) in zip(eachcol(weights), center_distances)
+        )
+    inner_inertia = _inner_inertia(metric, data, centers, weights, fuzziness)
```
alyst, line 237:
```suggestion
    outer_inertia = _outer_inertia(metric, data, centers, assignments, fuzziness)
    inner_inertia = _inner_inertia(metric, data, centers, assignments, fuzziness)
```

src/clustering_quality.jl
jaksle: It was unintended, I made a bad merge.

src/clustering_quality.jl
```diff
@@ -0,0 +1,314 @@
...
+    d == dc || throw(DimensionMismatch("Inconsistent array dimensions for `data` and `centers`."))
+    (1 <= k <= n) || throw(ArgumentError("Number of clusters k must be from 1:n (n=$n), k=$k given."))
+    k >= 2 || throw(ArgumentError("Quality index not defined for the degenerated clustering with a single cluster."))
+    n == k && throw(ArgumentError("Quality index not defined for the degenerated clustering where each data point is its own cluster."))
+    for i in eachindex(assignments)
+        (assignments[i] in axes(centers, 2)) || throw(ArgumentError("Bad assignments[$i]=$(assignments[i]) is not a valid index for `data`."))
+    end
+    for i in 1:k
+        i ∉ assignments && @warn "Cluster number $(i) is empty. Clustering quality calculation may not be reliable."
+    end
```
alyst, line 73: This is *O(N*K)* check, i.e. very expensive. We can improve it to *O(N+K)* if we scan assignments once and fill a boolean mask of observed cluster indices. Filling the mask could be done in the loop over assignments that you have above.
But ATM I'm not sure we need an explicit check.
Referring to the previous discussion -- we should have CI test cases with the gaps in the cluster indices (I think you have just added those), but if the returned results make sense (correct value is returned or reasonable error is thrown), we don't need to check in advance.

src/clustering_quality.jl
```diff
@@ -0,0 +1,314 @@
...
+    )
+    d, n = size(data)
+    dc, k = size(centers)
+
+    d == dc || throw(DimensionMismatch("Inconsistent array dimensions for `data` and `centers`."))
+    (1 <= k <= n) || throw(ArgumentError("Number of clusters k must be from 1:n (n=$n), k=$k given."))
+    k >= 2 || throw(ArgumentError("Quality index not defined for the degenerated clustering with a single cluster."))
+    n == k && throw(ArgumentError("Quality index not defined for the degenerated clustering where each data point is its own cluster."))
+    for i in eachindex(assignments)
+        (assignments[i] in axes(centers, 2)) || throw(ArgumentError("Bad assignments[$i]=$(assignments[i]) is not a valid index for `data`."))
```
alyst, line 69:
```suggestion
    for (i, clu) in enumerate(assignments)
        (clu in axes(centers, 2)) || throw(ArgumentError("Invalid cluster index: assignments[$i]=$(clu)."))
```

src/clustering_quality.jl
jaksle: I've checked and alas, every alorithm reacts differently to empty clusters. We have the whole range: normal val, NaN, Inf. That's why I thought warning could be for the best.

src/clustering_quality.jl
```diff
@@ -0,0 +1,314 @@
...
+    d == dc || throw(DimensionMismatch("Inconsistent array dimensions for `data` and `centers`."))
+    (1 <= k <= n) || throw(ArgumentError("Number of clusters k must be from 1:n (n=$n), k=$k given."))
+    k >= 2 || throw(ArgumentError("Quality index not defined for the degenerated clustering with a single cluster."))
+    n == k && throw(ArgumentError("Quality index not defined for the degenerated clustering where each data point is its own cluster."))
+    for (i, clu) in enumerate(assignments)
+        (clu in axes(centers, 2)) || throw(ArgumentError("Invalid cluster index: assignments[$i]=$(clu)."))
+    end
+    for i in 1:k
+        i ∉ assignments && @warn "Cluster number $(i) is empty. Clustering quality calculation may not be reliable."
+    end
```
alyst, line 73: Ok, then something like this should be a less expensive check for the empty clusters (just a draft, I have not tried running it):

```suggestion
    seen_clusters = falses(k)
    for (i, clu) in enumerate(assignments)
        (clu in axes(centers, 2)) || throw(ArgumentError("Invalid cluster index: assignments[$i]=$(clu)."))
        seen_clusters[clu] = true
    end
    if !all(seen_clusters)
        empty_clu_ixs = findall(~seen_clusters)
        @warn "Empty cluster(s) detected: $(join(String.(empty_clu_ixs), ", ")). Clustering quality calculation may not be reliable."
    end
```

One problem we have ATM is that this check is not done when you call another variant of `clustering_quality()` method.
So what would be optimal is to add the gaps check to [`check_assignments()` method](https://github.com/JuliaStats/Clustering.jl/blob/83adafa3d3e6853472d56ea2449c4359bfb07d77/src/utils.jl#L80) via the new `empty_clusters = :warn` keyword option,
and them we can just call `check_assignments()` from here and other public methods.

jaksle: Only methods which use cluster centers seem to have a problem with empty clusters, so it might be fine to give warning only in this case.

alyst: Actually, the better solution would be just to reindex the clusters skipping the empty ones:
```
old2new = cumsum(seen_clusters)
centers = view(centers, :, seen_clusters) # or findall(seen_clusters), whichever is faster, or even centers[:, seen_clusters]
assignments = old2new[assignments]
```
Not much more expensive, and gives the correct results.
And we can keep reporting the empty clusters (without the "incorrect results" notice), because the clusterings with empty clusters don't sound like 100% good quality.

jaksle: So, do you think we should include it only in hard clustering quality interface? And add keyword option as you suggested above?

alyst: Does fuzzy suffers from empty clusters (zero weight columns) too? If yes, we can do the same for fuzzy
The keyword option referred to the check_assignments() utility function. But here we are implementing more complex logic to workaround empty clusters, not just warn, so in the end we would not be calling check_assignments(). No need to add such option for clustering_quality(), I think - we just warn unconditionally.

jaksle: Fuzzy quality measures are quite robust, as long as weights are non-negative they are completely fine.

alyst:
> You mentioned wanting to reshape documentation. It may be a good idea to divide "Evaluation & Validation" into 2 sections.

Yes, that's what I had in mind: separate intrinsic quality metrics and the comparison between the clusterings (cross-validation is one use case).
I am keeping it out of scope for this PR, because you have already done a lot (thank you!).

alyst:
@jaksle I think we're almost done with the PR. But it would be nice to improve the coverage.
E.g. https://app.codecov.io/gh/JuliaStats/Clustering.jl/pull/257/blob/src/clustering_quality.jl#L89 and few other lines show that some parameter combinations are not tested.
Could you please add those tests? I think we would be ready to go then.

jaksle:
@alyst Yes, of course. For `:silhouettes` I haven't tested them deliberately assuming that what is here is basically an additional API for `silhouettes` method tested elsewhere.

alyst:
@jaksle Yes, I think no comprehensive tests are required for silhouettes.
It is just to check that *clustering_quality(..., quality_index=:silhouettes)* calculates the correct index.
If *silhouettes()* API changes, this test will help to identify that we need to update *clustering_quality()* code that calls it.

And there is also Dunn method that is not covered in that *clustering_quality()* method variant.

jaksle:
I added testing of `:silhouettes` (and wasted an hour looking for mistake in my analytical formula -_-). I remember why I haven't tested `:dunn` in line 91, I think it is the same call as line 152.

alyst (approved):

alyst:
@jaksle I've updated the tests to fix the coverage (also, I've enhanced the exception logic -- it should throw a correct error message if the index name is valid, but not compatible with input arguments). Plus, I've reshuffled the docs, so that clustering comparison and quality evaluation are the two big subsections. I think this PR is ready to be merged -- but let me know if you have some questions/concerns about my recent changes. And thank you for all your hard work!

jaksle:
This looks nice. I think this is a very good clean-up. I also think the PR is ready to merge. Thank you too for all the corrections!

PR #270, opened 2024-01-19
Small erorrs in docs validate links

jaksle:
I've noticed two broken links. This corrects them.  Additionally, in the plots in the same sections titles overlap with x labels which makes them unintelligible. But, compiling the same docs on my machine does not replicate this behaviour. I have Plots v1.39.0.

EDIT. I remade this PR, I had problems with synchronising fork to the origin.


alyst:
Thanks! The changes look straightforward, but there are conflicts with the master. It looks like you still have some mix-ups in the commits of this PR: there are 95 commits, which is a bit too much for 3 changed lines.
I guess the easiest would be if you just create a new branch off the Clustering.jl master and add a single new commit with the fixes to the ref links.
And in general, I think it is a good practice, when working on a PR, to create a new feature branch in your local repository instead of committing directly to the *main* (*master* one).

Note that for #257 I've squashed all your commits together (since it was all about adding new code) and force-pushed it (to the PR branch in your repository, which happened to be *master*), so that the commit history in the Clustering.jl is cleaner.

alyst:
BTW, do you have a quick fix for the overlapping titles issue in the quality indices docs?
I guess it is just about increasing the vertical padding/margins of the subplots. Would be nice to have it in this PR.

PR #271, opened 2024-01-23
Small corrections in docs validate.md

jaksle:
Third attempt at making this PR. I made a new repository and commit from a branch this time.

Corrections in links + changing `nclusters` to `№_clu`. There is a function `nclusters` already.

Something is deeply wrong with `Plots` when it compiles official docs for validate.md. Not only it puts titles on labels, it also does not display some labels at all. I  cannot reproduce it on two different machines, so currently I have no idea how to fix it.

EDIT. Looking more closely maybe it renders those labels, but they do not into the plot frames. Maybe we can try updating `Plots` version in docs?

### DIFF
--- a/docs/source/validate.md
+++ b/docs/source/validate.md
@@ -214,11 +214,11 @@ X = hcat([4., 5.] .+ 0.4 * randn(2, 10),
          [9., -5.] .+ 0.4 * randn(2, 5),
          [-4., -9.] .+ 1 * randn(2, 5))
 
-nclusters = 2:5
-clusterings = kmeans.(Ref(X), nclusters)
+№_clu = 2:5
+clusterings = kmeans.(Ref(X), №_clu)
 
 plot((
-    plot(nclusters,
+    plot(№_clu,
          clustering_quality.(Ref(X), clusterings, quality_index = qidx),
          marker = :circle,
          title = ":$qidx", label = nothing,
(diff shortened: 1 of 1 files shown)


docs/source/validate.md
```diff
@@ -214,11 +214,11 @@ X = hcat([4., 5.] .+ 0.4 * randn(2, 10),
          [9., -5.] .+ 0.4 * randn(2, 5),
          [-4., -9.] .+ 1 * randn(2, 5))
 
-nclusters = 2:5
-clusterings = kmeans.(Ref(X), nclusters)
+№_clu = 2:5
```
alyst, line 217: I think "№" symbol [typically denotes "sequential number"](https://en.wikipedia.org/wiki/Numero_sign), e.g. "cluster №3", but not the quantity of items.

jaksle: I am no expert here, I always thought it is just old abbreviation for "number" with no further nuance. We can change to `n_clu` or something similar.

alyst: I reverted it back to *nclusters*

alyst:
I have tried to tweak the layout of the plots and switched to SVG, let's see if that helps.

alyst:
The updates did fix the dev docs.

jaksle:
Thanks!

Issue #139, opened 2019-01-27
add HDBSCAN?


currymj:
DBSCAN is already included. There is a successor, [hdbscan](https://github.com/scikit-learn-contrib/hdbscan) which has a famously good Python package, and is fairly popular.

DBSCAN is already here, and there are hierarchical clustering algorithms as well, so it's possible some code could be reused. There's a good explanation [here](https://hdbscan.readthedocs.io/en/latest/how_hdbscan_works.html) of all the pieces of the algorithm.

I wish I were submitting a PR instead of just a feature request issue, but I still think a pure Julia implementation would be good to have.

Also, if anybody Googling for a Julia HDBSCAN implementation stumbles on this issue, you can just use [PyCall.jl](https://github.com/JuliaPy/PyCall.jl) to call the hdbscan Python package. It works fine, just remember to transpose your data matrix because the Python convention is the opposite of Julia.

baggepinnen:
I created a minimum-effort wrapper here
https://github.com/baggepinnen/HDBSCAN.jl

babaq:
a Julia version is always the best, but thanks for the wrapper @baggepinnen .

MommaWatasu:
I wrote the code only for learning and I haven't maintained it for long (since no one uses it).
I created PR which contains my code from HorseML.jl. I hope my code is useful.

PR #273, opened 2024-03-22
Add HDBSCAN from HorseML.jl

MommaWatasu:
# Add HDBSCAN
I found [this issue](https://github.com/JuliaStats/Clustering.jl/issues/139). I wanted to my code to be useful, so compiled it into hdbscan.jl so it works as is.
I changed some code from my original to get closer to the code of this repo.

## Abstract of functions and structures
- HDBSCANGraph: is used to build a minimum-spanning-tree
- HDBSCANCluster: is used to build cluster based on minimum-spanning-tree
- HDBSCANResult: is used to return the result
- hdbscan!: main function that performs hdbscan

As I wrote in comment, many utility functions are just converted from numpy by myself, so I don't know many about them.

## Usage
This is the usage of main function.

    hdbscan!(points::AbstractMatrix, k::Int64, min_cluster_size::INt64; gen_mst::Bool=true, mst=nothing)
### Parameters
- `points`: the d×n matrix, where each column is a d-dimensional coordinate of a point
- `k`: we will define "core distance of point A" as the distance between point A and the `k` th neighbor point of point A.
- `min_cluster_size`: minimum number of points in the cluster
- `gen_mst`: whether to generate minimum-spannig-tree or not
- `mst`: when is specified and `gen_mst` is false, new mst won't be generated

## Example
I checked that this following code is available:
```
# include hdbscan.jl before run this code
using CSV
using DataFrames
using Plots
data = CSV.read("/home/watasu/Documents/code/HorseML.jl/test/data/clustering2.csv", DataFrame) |> Matrix
result = hdbscan!(data, 5, 3)
plot(title = "Clustering by HDBSCAN")
result = result.labels
for i in -1 : maximum(result)
    X = data[findall(result.==i), :]
    plot!(X[:, 1], X[:, 2], st=:scatter)
end
plot!()
```


alyst:
@MommaWatasu Thanks for the PR! I think it would be a useful addition to the package. I will try to review it soon. Meanwhile -- it looks like there are no unit tests for the method. Could you please add some?

MommaWatasu:
I added some test for hdbscan. But I'm sorry for not being able to write unit test for utility functions.
If you want to write tests for them, you should check scipy for their specification.
They are coming from:
- erf: scipy.special.erf
- logpdf: I think the original code is [this](https://github.com/scipy/scipy/blob/38919c0ec0a9fdc27b4a720dfeba2ae4e56f20d6/scipy/stats/_multivariate.py#L537)

alyst:
@MommaWatasu Thanks for adding unit tests. Generally we don't test internal utility functions, we only test public API, but aim to cover both meaningful examples (small datasets with nontrivial clusterings) and corner cases (e.g. single data point).
As for some utility functions - erf/erfc are available in Distributions.jl, which Clusterings.jl already depends upon, so you should rather use these ones. pdf/logpdf are also declared there.

MommaWatasu:
@alyst I looked at Project.toml but Distributions.jl wasn't there. Moreover, it seems like it only has `logpdf` but not `erf`(at least, it's not exported).
Please tell me where did you found `erf` in Distributions.jl.

alyst:
> I noticed that erf and logpdf aren't for HDBSCAN (but for Xmeans). I deleted them and updated test. 

Great, *erf* is provided by *SpecialFunctions.jl*, but as we want to be conservative on the number of package dependencies, it is convenient that we don't need it.

src/hdbscan.jl
```diff
@@ -0,0 +1,296 @@
...
+    end
+end
+
+# Below are utility functions for building hierarchical trees
+heappush!(h, v) = insert!(h, searchsortedfirst(h, v), v)
+
+mutable struct UnionFind{T <: Integer}
+    parent:: Vector{T}  # parent[root] is the negative of the size
+    label::Dict{Int, Int}
+    cnt::Int
```
alyst, line 245: Can you please put some comments that describe what this function does and what are its fields?
I have not tried too hard to understand what it, but ideally that should be clear from the comments.
The only field that could be modified is `cnt`, but IIUC `length(uf.label)` gets you the same result, so you can just remove `cnt` and turn the type into `struct`.

```suggestion
struct UnionFind{T <: Integer}
    parent:: Vector{T}  # parent[root] is the negative of the size
    label::Dict{Int, Int}
```

test/hdbscan.jl
```diff
@@ -0,0 +1,19 @@
...
+    lower_y = cos.(lower_x) .+ rand(51)./10
+    data = hcat([lower_x; upper_x], [lower_y; upper_y])
+    #test for main function
+    @test_throws DomainError hdbscan(data, 5, 0)
+    @test_throws ArgumentError hdbscan(data, 5, 3; gen_mst=false)
+    @test_nowarn result = hdbscan(data, 5, 3)
+
+    # tests for result
+    result = hdbscan(data, 5, 3)
+    @test sum([length(c) for c in result.clusters]) == 101
```
alyst, line 18: We need more tests to make sure the results are correct.
In particular that the number of clusters is correct and that the points are assigned to the correct clusters.

```suggestion
    result = @inferred(hdbscan(data, 5, 3))
    @test results is HdbscanResult
    @test sum(length, result.clusters) == 101
```

alyst (changes requested):
Thanks again for the PR!
The first iteration looks good -- we don't need much refactoring, mostly some method renames and object field tweaks for clarity.

Of a bigger things:
  * check the dimensions of the `points` matrix
  * use `Distances.jl` API to calculate the point distances (see the other methods, e.g. `clustering_quality()` for how we do it); it is worth precalculating all pairwise distances and passing it to the methods that calculate core distances and build the graph. We can also allow the user to specify the metric as a kwarg to `hdbscan`, which would be a useful generalization.
  * add more tests to check that the point assignments to the clusters are correct

docs/source/hdbscan.md
```diff
@@ -0,0 +1,11 @@
+# HDBSCAN
```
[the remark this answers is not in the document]
MommaWatasu: I think it would. I didn't want to add a new file, but rather separate the files to differentiate my written document from other documents.

src/hdbscan.jl
```diff
@@ -0,0 +1,296 @@
+struct HDBSCANGraph
+    edges::Vector{Vector{Tuple{Int, Float64}}}
```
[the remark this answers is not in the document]
MommaWatasu: Yes, your guess is right. I applied your suggestion.

src/hdbscan.jl
```diff
@@ -0,0 +1,296 @@
...
+        for k in c.points
+            noise_points[k] = 0
+        end
+    end
+    push!(result, HdbscanCluster(true, Int[]))
+    result[end].points = findall(x->x==-1, noise_points)
+    return HdbscanResult(k, min_cluster_size, result)
+end
+
+function core_dist(points, k)
```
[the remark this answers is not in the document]
MommaWatasu: @alyst 
I changed code with Distances.jl like this:
`dists = vec(pairwise(SqEuclidean(), points[i:i, :], points; dims=1))`
But it seems to be harmful. Look at the following benchmark:
```
julia> @benchmark core_distances(points, 5)
BenchmarkTools.Trial: 100 samples with 1 evaluation.
 Range (min … max):  49.296 ms … 64.033 ms  ┊ GC (min … max): 2.63% … 2.00%
 Time  (median):     49.839 ms              ┊ GC (median):    3.05%
 Time  (mean ± σ):   50.050 ms ±  1.493 ms  ┊ GC (mean ± σ):  2.96% ± 0.35%

     ▅▂█▃▇▂ ▃▃                                                 
  ▃▄██████████▅▃▅▁▁▁▃▃▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃ ▃
  49.3 ms         Histogram: frequency by time        53.7 ms <

 Memory estimate: 198.36 MiB, allocs estimate: 5501.

julia> @benchmark core_dist(points, 5)
BenchmarkTools.Trial: 161 samples with 1 evaluation.
 Range (min … max):  29.093 ms … 49.369 ms  ┊ GC (min … max): 4.43% … 6.92%
 Time  (median):     29.666 ms              ┊ GC (median):    5.43%
 Time  (mean ± σ):   31.224 ms ±  3.439 ms  ┊ GC (mean ± σ):  5.11% ± 0.72%

  ▃█▃▂                                                         
  ████▄▂▃▂▂▁▃▁▁▂▃▂▁▃▁▁▁▁▁▁▁▂▁▂▄▂▂▃▃▃▃▂▁▁▃▁▃▁▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▂ ▂
  29.1 ms         Histogram: frequency by time        42.4 ms <

 Memory estimate: 195.76 MiB, allocs estimate: 7001.

```
`points` is a 500x100 random matrix. I don't know well about optimization, so I wonder why the performance got worse. which should I use the new code or old code?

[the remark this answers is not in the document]
alyst: What I have originally meant is to call `dists = pairwise(Euclidean(), points; dims=2)` once to calculate the distances between all pairs of points (see the [Distances.jl readme](https://github.com/JuliaStats/Distances.jl?tab=readme-ov-file#computing-pairwise-distances)).
I think this have to be done outside the loop, and maybe even before you call `core_dists()`, because you also need the distances in `hdbscan_graph()`.
Then inside the loop, you just take iterate over the columns of dists (e.g. `for (i, i_dists) in enumeate(eachcol(dists))`) to select the k smallest distances.

Note that if you do it once, and only pass `dists` matrix to `core_distances()` and `hdbscan_graph()`, you can allow the user to specify any other metric as a parameter to `hdbscan(...; metric = SqEuclidean())` (and then you call `dists = pairwise(metric, points, dims=2)`.

src/hdbscan.jl
```diff
@@ -0,0 +1,296 @@
...
+        nc1, nc2 = isnoise(Hierarchy[c1]), isnoise(Hierarchy[c2])
+        if !(nc1 || nc2)
+            #compute stability
+            compute_stability(Hierarchy[c1], λ)
+            compute_stability(Hierarchy[c2], λ)
+            #unite cluster
+            unite!(uf, j, k)
+            #create parent cluster
+            points = members(uf, group(uf, j))
+            Hierarchy[n+i] = HdbscanCluster(false, points)
```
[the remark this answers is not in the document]
MommaWatasu: I think this suggestion changes the behavior of function and causes an error.
Please tell me why do you suggest to use `push!`? Isn't it faster to create an array at first and assign values later?

src/hdbscan.jl
MommaWatasu: I found this isn't right. The `cnt` field is used to count the number of unite and determine the new group id.
So, I changed the field name into `next_id` for clarity.

MommaWatasu:
@alyst 
I applied all your suggestions. Is there anything else I have to fix?

src/hdbscan.jl
```diff
@@ -0,0 +1,311 @@
...
+    return HdbscanResult(result, assignments)
+end
+
+# calculate the core distances of the points
+function core_distances(dists::AbstractMatrix, k::Integer)
+    core_dists = Array{Float64}(undef, size(dists, 1))
+    for i in axes(dists, 1)
+        dist = sort(dists[i, :])
+        core_dists[i] = dist[k]
+    end
```
alyst, line 124: It just occurred to me that it could be optimized into
```suggestion
    for i in axes(dists, 2)
        core_dists[i] = partialsort!(dists[:, i], k)
    end
```

Note that in Julia iterating over columns is much more efficient than by rows, because one column occupies a continuous block of memory, whereas in the row the neighboring elements are stored *N* elements (the number of columns) apart.

src/hdbscan.jl
```diff
@@ -0,0 +1,311 @@
...
+    result[end].points = findall(x->x==-1, noise_points)
+    assignments = Array{Int}(undef, length(points))
+    for i in 1 : length(result)-1
+        assignments[result[i].points] = i
+    end
+    assignments[result[end].points] .= -1
+    return HdbscanResult(result, assignments)
+end
+
+# calculate the core distances of the points
```
alyst, line 118:
```suggestion
# calculate the core (k-th nearest) distances of the points
```

src/hdbscan.jl
```diff
@@ -0,0 +1,310 @@
...
+        end
+    end
+    push!(result, HdbscanCluster(Int[]))
+    result[end].points = findall(x->x==-1, noise_points)
+    assignments = Array{Int}(undef, length(points))
+    for i in 1 : length(result)-1
+        assignments[result[i].points] = i
+    end
+    assignments[result[end].points] .= -1
+    return HdbscanResult(result, assignments)
```
alyst, line 115:
```suggestion
    clusters = HdbscanCluster[]
    assignments = fill(0, length(points)) # cluster index of each point
    for (i, j) in enumerate(hierarchy[2n-1].children)
        clu = hierarchy[j]
        push!(clusters, clu)
        assignments[clu.points] .= length(clusters)
    end
    # add the cluster of all unassigned (noise) points
    push!(clusters, HdbscanCluster(findall(==(0), assignments)))
    return HdbscanResult(clusters, assignments)
```

src/hdbscan.jl
```diff
@@ -0,0 +1,303 @@
...
+    v1::Integer
+    v2::Integer
+    dist::Number
+end
+expand(edge::MSTEdge) = (edge.v1, edge.v2, edge.dist)
+Base.isless(edge1::MSTEdge, edge2::MSTEdge) = edge1.dist < edge2.dist
+
+"""
+    HdbscanCluster
+A cluster generated by the [hdbscan](@ref) method, part of [HdbscanResult](@ref).
```
alyst, line 27:
```suggestion
    HdbscanCluster

A cluster generated by the [hdbscan](@ref) method, part of [HdbscanResult](@ref).
```

@MommaWatasu Please make sure that in all your docstrings there is an empty line between the declaration of the structure or a method and its description. That's the [docstring requirement](https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation).

src/hdbscan.jl
```diff
@@ -0,0 +1,303 @@
+# HDBSCAN Graph
+# edge[i] is a list of edges adjacent to the i-th vertex
+# the second element of HDBSCANEdge is the mutual reachability distance.
+HDBSCANEdge = Tuple{Int, Float64}
+struct HDBSCANGraph
+    edges::Vector{Vector{HDBSCANEdge}}
+    HDBSCANGraph(nv::Integer) = new([HDBSCANEdge[] for _ in 1 : nv])
```
alyst, line 7:
```suggestion
    adj_edges::Vector{Vector{HDBSCANEdge}} # adj_edges[i] is a vector of edges adjacent to the i-th vertex
  
    HDBSCANGraph(nv::Integer) = new([HDBSCANEdge[] for _ in 1 : nv])
```

Please add the newline between the field definitions and the inner ctor in all your structure declarations.
I suggest to rename `graph.edges` into `graph.adj_edges`, because if one just read the field name `edges` it could be understood as a list of all edges in the graph.

src/hdbscan.jl
```diff
@@ -0,0 +1,303 @@
+# HDBSCAN Graph
+# edge[i] is a list of edges adjacent to the i-th vertex
+# the second element of HDBSCANEdge is the mutual reachability distance.
+HDBSCANEdge = Tuple{Int, Float64}
+struct HDBSCANGraph
```
alyst, line 5:
```suggestion
# HDBSCAN Graph edge: target vertex and mutual reachability distance.
HDBSCANEdge = Tuple{Int, Float64}

# HDBSCAN Graph
struct HDBSCANGraph
```

src/hdbscan.jl
```diff
@@ -0,0 +1,303 @@
...
+    HDBSCANGraph(nv::Integer) = new([HDBSCANEdge[] for _ in 1 : nv])
+end
+
+function add_edge!(G::HDBSCANGraph, v1::Integer, v2::Integer, dist::Number)
+    push!(G.edges[v1], (v2, dist))
+    push!(G.edges[v2], (v1, dist))
+end
+
+Base.getindex(G::HDBSCANGraph, i::Int) = G.edges[i]
+
```
alyst, line 16: `HDBSCANGraph` is an internal structure, so we don't need to overload Base API just for that.
But also I think `graph[i]` does not help to make the code better readable -- it is not clear, which objects you are accessing. I think `graph.adj_edges[i]` is better.

```suggestion
```

src/hdbscan.jl
```diff
@@ -0,0 +1,303 @@
...
+# Below are utility functions for building hierarchical trees
+heappush!(h, v) = insert!(h, searchsortedfirst(h, v), v)
+
+# Union-Find
+# structure for managing disjoint sets
+# This structure tracks which sets the elements of a set belong to,
+# and allows us to efficiently determine whether two elements belong to the same set.
+mutable struct UnionFind
+    parent:: Vector{Integer}  # parent[root] is the negative of the size
```
alyst, line 246: `Integer` is an abstract type, so the performance of `Vector{Integer}` is suboptimal, because Julia has to do runtime type inference for each access to its elements.

```suggestion
    parent::Vector{Int}  # parent[root] is the negative of the size
```

src/hdbscan.jl
```diff
@@ -0,0 +1,303 @@
...
+    label::Dict{Int, Int}
+    next_id::Int
+
+    function UnionFind(nodes::Integer)
+        if nodes <= 0
+            throw(ArgumentError("invalid argument for nodes: $nodes"))
+        end
+
+        parent = -ones(nodes)
+        label = Dict([(i, i) for i in 1 : nodes])
```
alyst, line 256:
```suggestion
        label = Dict(i => i for i in 1:nodes)
```

src/hdbscan.jl
```diff
@@ -0,0 +1,303 @@
...
+        new(parent, label, nodes)
+    end
+end
+
+# label of the set which element `x` belong to
+group(uf::UnionFind, x) = uf.label[root(uf, x)]
+# all elements that have the specified label
+members(uf::UnionFind, x::Int) = collect(keys(filter(n->n.second == x, uf.label)))
```
alyst, line 264:
```suggestion
members(uf::UnionFind, x::Integer) = [k for (k, v) in pairs(uf.label) if v == x]
```

src/hdbscan.jl
```diff
@@ -0,0 +1,303 @@
...
+function root(uf::UnionFind, x::Integer)
+    if uf.parent[x] < 0
+        return x
+    else
+        return uf.parent[x] = root(uf, uf.parent[x])
+    end
+end
+
+# whether element `x` and `y` belong to the same set
+function issame(uf::UnionFind, x::Integer, y::Integer)
```
alyst, line 277:
```suggestion
function issamegroup(uf::UnionFind, x::Integer, y::Integer)
```

src/hdbscan.jl
```diff
@@ -0,0 +1,303 @@
...
+# The root is the representative element of the set
+function root(uf::UnionFind, x::Integer)
+    if uf.parent[x] < 0
+        return x
+    else
+        return uf.parent[x] = root(uf, uf.parent[x])
+    end
+end
+
+# whether element `x` and `y` belong to the same set
```
alyst, line 276: Let's use the same naming convention in the code and in the comments.
Now you API uses `group`/`members` terms, while the comments refer to `sets`.
I think I would prefer `set_id()` and `items()`

```suggestion
# whether element `x` and `y` belong to the same group
```

src/hdbscan.jl
```diff
@@ -0,0 +1,303 @@
...
+function issame(uf::UnionFind, x::Integer, y::Integer)
+    return root(uf, x) == root(uf, y)
+end
+
+function Base.size(uf::UnionFind, x::Integer)
+    return -uf.parent[root(uf, x)]
+end
```
alyst, line 283: That overload does not conform to `Base.size(obj, i)` API convention, which should give the size of `i`-th dimension of an array. This has to be absolutely avoided. Plus, we don't need to overload Base API for internal types.

It's not 100% clear what this function does, if it returns the set size, let's just call it this way:
```suggestion
function setsize(uf::UnionFind, x::Integer)
    return -uf.parent[root(uf, x)]
end
```

alyst:
> I applied all your suggestions. Is there anything else I have to fix?

@MommaWatasu Thank you for adjusting the PR! We are getting close to be able to merge, but we would need one more iteration. 

Note that I have pushed some adjustments to the code directly to your branch, so please make sure to pull them first.

TODO items:
  * at the moment the tests that you have do not really test whether the clusters are correct. In fact, at the moment all the points are assigned to the noise cluster (I think it was also the case before my changes).
    Please add the test(s) that the assignments/clusters are correct. Ideally, we need to test a more complex clustering (more than 2 clusters), also would be nice to test that changing *ncore*, or *min_cluster_size* affects the result
  * Now that I understand the algorithm a bit more, it looks like `HdbscanCluster` is an internal structure, and it is rather a `HdbscanTree` node than the cluster you return to the user. It contains the fields like `stability`, `children` etc: some of them we should not expose to the user, the others you don't really set when you are preparing the resulting clusters. I suggest that you rename this structure into `HdbscanNode` (this is a non-exported structure for the algorithm), and create a new one, `HdbscanCluster` (the exported one returned to the user). If there are any properties of the cluster, such as `stability` or being noise -- please make sure to add these relevant fields to the `HdbscanCluster` and properly initialize them when you are generating the result.
  * `HdbscanResult` should inherit from `ClusteringResult` and support its API (in particular, counts is missing)
  * move `UnionFind` to a separate source file *unionfind.jl*. I'm not 100% sure it belongs to *Clustering.jl*. We may potentially use the *DataStructures.jl*, but this code looks rather compact, so I think I prefer keeping it over depending on another big package.
  * cleanup `UnionFind` terminology. I think `set`/`issameset`/`items` would be an optimal choice.
  * add the unit tests for `UnionFind`, e.g. that finding root, issameset, unite! work as expected
  * add the newlines in the docstrings that separate declaration from the description
  * add newlines that separate struct fields from the inner constructor

MommaWatasu:
@alyst I have one thing to apology. I found that the reason why the clustering result went wrong was my serious mistake about the algorithm. I fixed the algorithm and checked the result is correct. In addition, all the TODO items have been done(but I couldn't add the unit test about `ncore` because I don't know how it effects to the result).

I would appreciate it if you could check for any performance issues regarding the fixed algorithm.

sztal:
Hi, thanks for the effort of bringing HDBSCAN to Clustering.jl! I wonder what is the current status of this issue?

BenCurran98:
I would also be keen to get this merged soon, it looks great :)

src/hdbscan.jl
```diff
@@ -0,0 +1,232 @@
...
+- `metric`(defaults to Euclidean): the points distance metric to use.
+"""
+function hdbscan(points::AbstractMatrix, ncore::Integer, min_cluster_size::Int; metric=Euclidean())
+    if min_cluster_size < 1
+        throw(DomainError(min_cluster_size, "The `min_cluster_size` must be greater than or equal to 1"))
+    end
+    n = size(points, 2)
+    dists = pairwise(metric, points; dims=2)
+    # calculate core (ncore-th nearest) distance for each point
+    core_dists = [partialsort!(dists[:, i], ncore) for i in axes(dists, 2)]
```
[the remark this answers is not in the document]
MommaWatasu: `partialsort!` doesn't return a vector but the k-th value. So, I think this is correct.

BenCurran98:
@alyst Any updates on this one?

PR #283, opened 2024-09-07
Fix broadcasting bug in `repick_unused_centers()` for K-means

AzamatB:


alyst:
@AzamatB Thank you for the fix and sorry for very long delay in reviewing it! Do you have a test case for the fix?

alyst:
It is clearly a bug, it affects the weights of points during empty clusters reinitialization.
But from what I can understand, it should not significantly affect the correctness of the results.
In some rare cases it can lead to the same point becoming the center of two unused clusters -- but it is hard to construct this case.
We should definitely fix it, but it would be nice to have a test case that reveals the problematic behaviour.
So if you discovered this bug because kmeans() generated wrong results for your data, it would be nice to have that example.

Issue #280, opened 2024-06-04
Unnecessary check of dimensions for input?


BioTurboNick:
Is this check necessary?

Would be nicer to emit a trivial clustering rather than to error.

https://github.com/JuliaStats/Clustering.jl/blob/74d5d43a2cf4f3d40a6e0d9b7c820d169f2f9aaa/src/dbscan.jl#L104

lwabeke:
Bump on this, I don't have a lot of experience with DBSCAN, but I don't understand the point of this exception.
From what I can determine, the code should excecute without this check.
Not sure if I would call clustering any more trivial in this case, consider that expanding dim from 1 to 100 does not change the original inherent problem of deciding if the 2 points are close together or not:
```
dim = 100
points = zeros(dim)
points[1,:] .= 1.0
clustersTogether = dbscan(points, 1)

points[1,2] = 10
clustersApart  = dbscan(points, 1)
length(clustersTogether.counts) != length(clustersApart.counts)

```

BioTurboNick:
I was assuming that the error was added because incorrect results can be produced if the number of dimensions is greater than the number of points present. So by "trivial clustering" I meant "each point is a cluster". or an empty set if `min_cluster_size` is greater than 1.

alyst:
Thank you for the report!
Can you submit the PR with the fix and the tests case (D >= N) confirming that the algorithm works correctly?

PR #286, opened 2025-01-03
Modification to dbscan to add test cases and support for the case where num points <= num dimensions

lwabeke:
This relates to #280


alyst:
Thank you! Can you also add the test for 0 points?

